PR #2, opened 2023-03-28
Changes to ROS2 Wrapper to Combine with ROS1 Wrapper

jbarry-bdai:
This is a review representing the set of changes made to the spot ros2 wrapper by replacing it with this repo (not quite sure what the right way to give feedback on this was but this seemed reasonable).

- @METEORITENMAX your team should look through this carefully.  I focused primarily on the functionality my team is using right now, which is the basic stuff plus robot command - I did not check carefully the graph nav integration etc.
- @DavidWatkins Can you check the image publishing in light of the changes you made in https://github.com/bdaiinstitute/spot_ros2/pull/23/files#diff-6a84671d329adaf336fe25d35ef2fd6c75ff3ca5e64210cb6e469120404cdae0
- @heuristicus I've already left a few questions and requests for changes if you want to look at those.  I can also make them directly and put up a real PR for them - let me know your preferences.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -1,39 +1,92 @@
+import functools
+import logging
 import math
 import time
 import traceback
+import typing
 from collections import namedtuple
-from typing import Optional
 
-import cv2
-import numpy as np
-from bosdyn.client import create_standard_sdk, ResponseError, RpcError
-from bosdyn.client.async_tasks import AsyncPeriodicQuery, AsyncTasks
-
-from bosdyn.client.robot_state import RobotStateClient
-from bosdyn.client.robot_command import RobotCommandClient, RobotCommandBuilder
-from bosdyn.client.graph_nav import GraphNavClient
-from bosdyn.client.frame_helpers import get_odom_tform_body
-from bosdyn.client.power import safe_power_off, PowerClient, power_on
-from bosdyn.client.lease import LeaseClient, LeaseKeepAlive
-from bosdyn.client.image import ImageClient, build_image_request, UnsupportedPixelFormatRequestedError
+import bosdyn.client.auth
+from bosdyn.api import arm_command_pb2
+from bosdyn.api import geometry_pb2
 from bosdyn.api import image_pb2
+from bosdyn.api import manipulation_api_pb2
+from bosdyn.api import robot_command_pb2
+from bosdyn.api import robot_state_pb2
+from bosdyn.api import synchronized_command_pb2
+from bosdyn.api import trajectory_pb2
 from bosdyn.api.graph_nav import graph_nav_pb2
 from bosdyn.api.graph_nav import map_pb2
 from bosdyn.api.graph_nav import nav_pb2
-from bosdyn.client.estop import EstopClient, EstopEndpoint, EstopKeepAlive, MotorsOnError
+from bosdyn.client import create_standard_sdk, ResponseError, RpcError
 from bosdyn.client import frame_helpers
 from bosdyn.client import math_helpers
+from bosdyn.client import robot_command
+from bosdyn.client.async_tasks import AsyncPeriodicQuery, AsyncTasks
+from bosdyn.client.docking import DockingClient, blocking_dock_robot, blocking_undock
+from bosdyn.client.estop import (
+    EstopClient,
+    EstopEndpoint,
+    EstopKeepAlive,
+    MotorsOnError,
+)
+from bosdyn.client.frame_helpers import get_odom_tform_body
+from bosdyn.client.graph_nav import GraphNavClient
+from bosdyn.client.image import (
+    ImageClient,
+    build_image_request,
+    UnsupportedPixelFormatRequestedError,
+)
+from bosdyn.client.lease import LeaseClient, LeaseKeepAlive
+from bosdyn.client.manipulation_api_client import ManipulationApiClient
+from bosdyn.client.point_cloud import build_pc_request
+from bosdyn.client.power import safe_power_off, PowerClient, power_on
+from bosdyn.client.robot_command import RobotCommandClient, RobotCommandBuilder
+from bosdyn.client.robot_state import RobotStateClient
 from bosdyn.client.world_object import WorldObjectClient
+from bosdyn.geometry import EulerZXY
+from bosdyn.util import seconds_to_duration
+from google.protobuf.duration_pb2 import Duration
 
 MAX_COMMAND_DURATION = 1e5
 
 ### Release
 from . import graph_nav_util
+
 ### Debug
 # import graph_nav_util
 
 from bosdyn.api import basic_command_pb2
 from google.protobuf.timestamp_pb2 import Timestamp
+
+front_image_sources = [
+    "frontleft_fisheye_image",
+    "frontright_fisheye_image",
+    "frontleft_depth",
+    "frontright_depth",
+]
+"""List of image sources for front image periodic query"""
+side_image_sources = [
+    "left_fisheye_image",
+    "right_fisheye_image",
+    "left_depth",
+    "right_depth",
+]
+"""List of image sources for side image periodic query"""
+rear_image_sources = ["back_fisheye_image", "back_depth"]
+"""List of image sources for rear image periodic query"""
+VELODYNE_SERVICE_NAME = "velodyne-point-cloud"
+"""Service name for getting pointcloud of VLP16 connected to Spot Core"""
+point_cloud_sources = ["velodyne-point-cloud"]
+"""List of point cloud sources"""
+hand_image_sources = [
+    "hand_image",
+    "hand_depth",
+    "hand_color_image",
+    "hand_depth_in_hand_color_frame",
+]
+"""List of image sources for hand image periodic query"""
+
 
 # TODO: Missing Hand images
 CAMERA_IMAGE_SOURCES = [
@@ -41,7 +94,7 @@ CAMERA_IMAGE_SOURCES = [
     "frontright_fisheye_image",
     "left_fisheye_image",
     "right_fisheye_image",
-    "back_fisheye_image"
+    "back_fisheye_image",
 ]
 DEPTH_IMAGE_SOURCES = [
     "frontleft_depth",
@@ -57,22 +110,25 @@ DEPTH_REGISTERED_IMAGE_SOURCES = [
     "left_depth_in_visual_frame",
     "back_depth_in_visual_frame",
 ]
-ImageBundle = namedtuple("ImageBundle", ["frontleft", "frontright", "left", "right", "back"])
+ImageBundle = namedtuple(
+    "ImageBundle", ["frontleft", "frontright", "left", "right", "back"]
+)
 
 
 class AsyncRobotState(AsyncPeriodicQuery):
     """Class to get robot state at regular intervals.  get_robot_state_async query sent to the robot at every tick.  Callback registered to defined callback function.
 
-        Attributes:
-            client: The Client to a service on the robot
-            logger: Logger object
-            rate: Rate (Hz) to trigger the query
-            callback: Callback function to call when the results of the query are available
+    Attributes:
+        client: The Client to a service on the robot
+        logger: Logger object
+        rate: Rate (Hz) to trigger the query
+        callback: Callback function to call when the results of the query are available
     """
 
     def __init__(self, client, logger, rate, callback):
-        super(AsyncRobotState, self).__init__("robot-state", client, logger,
-                                              period_sec=1.0 / max(rate, 1.0))
+        super(AsyncRobotState, self).__init__(
+            "robot-state", client, logger, period_sec=1.0 / max(rate, 1.0)
+        )
         self._callback = None
         if rate > 0.0:
             self._callback = callback
@@ -87,16 +143,17 @@ class AsyncRobotState(AsyncPeriodicQuery):
 class AsyncMetrics(AsyncPeriodicQuery):
     """Class to get robot metrics at regular intervals.  get_robot_metrics_async query sent to the robot at every tick.  Callback registered to defined callback function.
 
-        Attributes:
-            client: The Client to a service on the robot
-            logger: Logger object
-            rate: Rate (Hz) to trigger the query
-            callback: Callback function to call when the results of the query are available
+    Attributes:
+        client: The Client to a service on the robot
+        logger: Logger object
+        rate: Rate (Hz) to trigger the query
+        callback: Callback function to call when the results of the query are available
     """
 
     def __init__(self, client, logger, rate, callback):
-        super(AsyncMetrics, self).__init__("robot-metrics", client, logger,
-                                           period_sec=1.0 / max(rate, 1.0))
+        super(AsyncMetrics, self).__init__(
+            "robot-metrics", client, logger, period_sec=1.0 / max(rate, 1.0)
+        )
         self._callback = None
         if rate > 0.0:
             self._callback = callback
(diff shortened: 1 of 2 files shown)


spot_wrapper/wrapper.py
```diff
@@ -259,25 +442,90 @@ def _start_query(self):
...
+    Returns:
+        Decorator which will wrap the decorated function
+    """
+    # If this decorator is being used without the power_on arg, return it as if it was called with that arg specified
+    if func is None:
+        return functools.partial(try_claim, power_on=power_on)
+
+    @functools.wraps(func)
+    def wrapper_try_claim(self, *args, **kwargs):
+        if self._get_lease_on_action:
```
jbarry-bdai, line 464: Note to Jenny: we will want to set `get_lease_on_action` true

spot_wrapper/wrapper.py
```diff
@@ -212,42 +313,124 @@ def _start_query(self):
...
+        elif (
+            last_estop_status
+            == self._spot_wrapper._estop_keepalive.KeepAliveStatus.DISABLED
+        ):
+            self._logger.error(
+                "Estop keepalive is disabled: {}".format(last_estop_status[1])
+            )
+        else:
+            # estop keepalive is ok
+            pass
```
jbarry-bdai, line 417: Note to Jenny: need to test this without the driver holding the estop

spot_wrapper/wrapper.py
```diff
@@ -601,90 +1030,132 @@ def _robot_command(self, command_proto, end_time_secs=None, timesync_endpoint=No
...
+    @try_claim(power_on=True)
     def self_right(self):
         """Have the robot self-right itself."""
-        response = self.power_on(allow_already_powered=True)
-        if not response[0]:
-            return response
         response = self._robot_command(RobotCommandBuilder.selfright_command())
         return response[0], response[1]
 
+    @try_claim
```
jbarry-bdai, line 1058:
```suggestion
    @try_claim(power_on=True)
```

spot_wrapper/wrapper.py
```diff
@@ -601,90 +1030,132 @@ def _robot_command(self, command_proto, end_time_secs=None, timesync_endpoint=No
...
+            rid = self._robot_command_client.clear_behavior_fault(
+                behavior_fault_id=id, lease=None
+            )
             return True, "Success", rid
         except Exception as e:
             return False, str(e), None
 
-    def power_on(self, allow_already_powered=False):
+    @try_claim
+    def power_on(self):
```
jbarry-bdai, line 1143: This removes the `allow_already_powered_on` flag in favor of always returning true if the robot was already powered on.  We weren't using the functionality of returning false if the robot was already powered on but heads up to @METEORITENMAX

spot_wrapper/wrapper.py
```diff
@@ -939,9 +1923,11 @@ def _upload_graph_and_snapshots(self, upload_filepath):
         if not localization_state.localization.waypoint_id:
             # The robot is not localized to the newly uploaded graph.
             self._logger.info(
-                "Upload complete! The robot is currently not localized to the map; please localize", \
-                "the robot using commands (2) or (3) before attempting a navigation command.")
+                "Upload complete! The robot is currently not localized to the map; please localize",
+                "the robot using commands (2) or (3) before attempting a navigation command.",
+            )
 
+    @try_claim
```
jbarry-bdai, line 1930: Should this have `try_claim` as a private function?

spot_wrapper/wrapper.py
```diff
@@ -739,47 +1227,48 @@ def trajectory_cmd(self, goal_x, goal_y, goal_heading, cmd_duration, frame_name=
...
-            raise ValueError('frame_name must be \'vision\' or \'odom\'')
+            raise ValueError("frame_name must be 'vision' or 'odom'")
         if response[0]:
             self._last_trajectory_command = response[2]
         return response[0], response[1]
 
-    def robot_command(self, robot_command):
-        end_time = time.time() + MAX_COMMAND_DURATION
-        return self._robot_command(robot_command, end_time_secs=end_time,
-                                   timesync_endpoint=self._robot.time_sync.endpoint)
```
jbarry-bdai, line 781: This function should be restored please - we use it extensively!

jbarry-bdai (commented):
This looks great!  Some comments.

spot_wrapper/wrapper.py
```diff
@@ -289,71 +537,157 @@ def __init__(self, username, password, hostname, robot_name, logger, start_estop
...
+
+        self._point_cloud_requests = []
+        for source in point_cloud_sources:
+            self._point_cloud_requests.append(build_pc_request(source))
+
+        self._hand_image_requests = []
+        for source in hand_image_sources:
+            self._hand_image_requests.append(
+                build_image_request(source, image_format=image_pb2.Image.FORMAT_RAW)
+            )
```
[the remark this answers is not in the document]
DavidWatkins: I would highly recommend moving away from using the original method of callbacks for image publishing. This adds a lot of unneeded latency.

spot_wrapper/wrapper.py
```diff
@@ -363,27 +697,98 @@ def __init__(self, username, password, hostname, robot_name, logger, start_estop
...
                 self._robot_metrics_task,
                 self._lease_task,
+                self._front_image_task,
                 self._idle_task,
+                self._estop_monitor,
                 self._world_objects_task,
-            ])
+            ]
+
+            if self._point_cloud_client:
```
DavidWatkins, line 774: We can use the build int point cloud service, but I would instead recommend using the ROS based one as the robot is quite slow at publishing that much data. Using depth_image_proc is much more performant.

spot_wrapper/wrapper.py
```diff
@@ -1127,48 +2163,142 @@ def _match_edge(self, current_edges, waypoint1, waypoint2):
...
+                        pixel_format=image_pb2.Image.PIXEL_FORMAT_RGB_U8,
+                        quality_percent=50,
+                    )
+                ]
+            )[0]
         except UnsupportedPixelFormatRequestedError as e:
             return None
 
-    def get_images(self, image_requests: list[image_pb2.ImageRequest]) -> Optional[ImageBundle]:
+    def get_images(
```
DavidWatkins, line 2299: Is your team using 80char for your style guide? We are using 120 which is why this wasn't broken onto multiple lines.

DavidWatkins (commented):
I would advise against publishing point clouds directly from the spot. I would also suggest not using the AsyncImageService to publish images. The way the driver is written now it will be publishing twice. Additionally this paradigm has no ability to turn off depth publishing which is where lots of latency stems from.

spot_wrapper/wrapper.py
heuristicus: I've instituted the use of the default black formatter https://black.readthedocs.io/en/stable/index.html for the ROS1 repository. The default line length is 88 (https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html#line-length).

spot_wrapper/wrapper.py
```diff
@@ -212,42 +313,124 @@ def _start_query(self):
...
+        # command can cause other commands to be interrupted before they get to start
+        if (
+            self._spot_wrapper.is_standing
+            and not self._spot_wrapper.is_moving
+            and self._spot_wrapper._last_trajectory_command is not None
+            and self._spot_wrapper._last_stand_command is not None
+            and self._spot_wrapper._last_velocity_command_time is not None
+            and self._spot_wrapper._last_docking_command is not None
+        ):
+            self._spot_wrapper.stand(False)
```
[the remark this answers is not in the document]
heuristicus: I have to admit that its purpose is not entirely clear to me either. The whole AsyncIdle looks like it could do with a review to see if it's actually necessary.

spot_wrapper/wrapper.py
```diff
@@ -212,42 +313,124 @@ def _start_query(self):
...
+
+
+class AsyncEStopMonitor(AsyncPeriodicQuery):
```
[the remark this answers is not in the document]
heuristicus: My original PR which added this is at https://github.com/heuristicus/spot_ros/pull/72. Part of the reason I added it was to prevent an error which often came up when killing the driver because of an uninitialised variable, but also there was no monitoring of the estop client and I thought it would be wise to include something which checked its status.

spot_wrapper/wrapper.py
heuristicus: In my PR for the ros2 repo I made sure to initialise the wrapper with that parameter set to true: https://github.com/heuristicus/spot_ros2/blob/866098eccebaac9ac509f918b08293af020fb2cb/spot_driver/spot_driver/spot_ros2.py#L1020-L1021.

spot_wrapper/wrapper.py
```diff
@@ -289,71 +537,157 @@ def __init__(self, username, password, hostname, robot_name, logger, start_estop
...
-            self._valid = False
-            return
-
-        self._logger.info("Establishing time sync with robot - this could take some time.")
-        self._robot.time_sync.wait_for_sync(10.0)
-
-        if self._robot:
-            # Clients
+        authenticated = False
+        while not authenticated:
```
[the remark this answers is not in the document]
heuristicus: I don't think it should, `time.sleep` can receive keyboard interrupts. I haven't specifically tested this but I've been using the driver with this code for a while and it's not been an issue. To test this the driver has to run before the robot is fully booted up. 

Part of the reason I added this is because the previous implementation meant that if you started the driver before all services and so on were available, it would just crash or maybe even fail silently rather than doing the smart thing and waiting until everything was available.

spot_wrapper/wrapper.py
heuristicus: I'm not sure I understand what you mean by a ROS based one. We have to retrieve the pointclouds using the SDK interaction before we can pass them to ROS. Perhaps you mean that we should split this out like you did with the image publishing?

spot_wrapper/wrapper.py
```diff
@@ -567,7 +992,11 @@ def releaseEStop(self):
 
     def getLease(self):
         """Get a lease for the robot and keep the lease alive automatically."""
-        self._lease = self._lease_client.take()
+        if self._use_take_lease:
+            self._lease = self._lease_client.take()
+        else:
+            self._lease = self._lease_client.acquire()
```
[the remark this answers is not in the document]
heuristicus: This is also set to true by default in the ros2 PR.

spot_wrapper/wrapper.py
heuristicus: I think I understand why this might be necessary - if the robot is on the floor but not in the correct sitting position (e.g. legs in bad orientation) and not powered on, this would be applicable. I was thinking that if you are trying to sit then you must by definition be standing, but that's not necessarily the case.

spot_wrapper/wrapper.py
heuristicus: Yes, the reason I removed this was because it didn't appear to be used anywhere in the ros2 driver.

spot_wrapper/wrapper.py
heuristicus: I removed this because I couldn't find any calls to this function within the wrapper or the ros2 driver code. Perhaps I missed something. Or do you mean that you are calling this from code that is external to the ros2/wrapper code?

spot_wrapper/wrapper.py
heuristicus: Can you clarify what you mean by having it as a private function? Do you mean an inner function in `_navigate_to`? Is the reason it is necessary for the case where toggling power fails?

spot_wrapper/wrapper.py
DavidWatkins: The package `depth_image_proc` here: http://wiki.ros.org/depth_image_proc. This package takes in an RGB image and a depth image and publishes the point cloud. This means the robot doesn't have to transfer the point cloud data but instead can just publish a depth and rgb image only. This can be added as a launch file which I already did in the `spot_ros2` repo.

DavidWatkins:
In our solution we actually didn't use a separate node to publish images. We just didn't rely on the services you created to do so. In our solution there is a synchronous call on a timer that gets the image and then publishes them. This plays nicer with how ROS2 handles threads. 

The point cloud publisher should be a separate node in the form of `depth_image_proc`. This would then accept RGB and Depth images, which are already being published, and convert them into a point cloud without using more bandwidth from the robot.

spot_wrapper/wrapper.py
heuristicus: Ah, I see that we're probably talking about two different things now. The point cloud client isn't used to publish the pointclouds from the cameras, but the one which comes from the VLP16 from the official EAP that BD provides.

spot_wrapper/wrapper.py
jbarry-bdai: It's on line 636 of `spot_ros2.py` on the main branch of the ROS2 code in `handle_robot_command`.
```
    def handle_robot_command(self, goal_handle):
        ros_command = goal_handle.request.command
        proto_command = robot_command_pb2.RobotCommand()
        conv.convert_bosdyn_msgs_robot_command_to_proto(ros_command, proto_command)
        self._wait_for_goal = None
        if not self.spot_wrapper:
            self._wait_for_goal = WaitForGoal(self.node.get_clock(), 2.0)
            goal_id = None
        else:
            success, err_msg, goal_id = self.spot_wrapper.robot_command(proto_command)
            if not success:
                raise Exception(err_msg)
```

Admittedly that's the only call to it - when I said "use it extensively" I mean we use the robot command action a lot not that this is called from multiple places :)

spot_wrapper/wrapper.py
jbarry-bdai: I meant that since `_navigate_to` was a private function should the supposition be that `try_claim` was already called.

spot_wrapper/wrapper.py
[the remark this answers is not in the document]
jbarry-bdai: OK my read of this is that so long as we don't actually create a `front_image` callback etc in our `spot_ros2`, this code can stay here without impacting anything.

@DavidWatkins can you tell me what to test to make sure we're still getting images at a good rate?  Just `rostopic hz`?

spot_wrapper/wrapper.py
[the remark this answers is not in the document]
jbarry-bdai: I'm just adding another parameter for it

spot_wrapper/wrapper.py
jbarry-bdai: Oh yeah I think I added that when I updated some of the power on stuff to preserve old behavior.  Since no one has needed it in the months since I did that I'm going to say it's fine to remove :)

spot_wrapper/wrapper.py
jbarry-bdai: I think it would be pretty easy to add this parameter but I'm not sure I really see the point of it - so long as the rate is set to negative or no callback is assigned, this just creates a client that isn't used, which doesn't seem like a big deal.

jbarry-bdai: Actually rather than creating the parameter, we'd just check if `lidar_points` existed in the callbacks list.

jbarry-bdai:
I'm working on some of these changes in #3 fyi.

spot_wrapper/wrapper.py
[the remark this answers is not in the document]
DavidWatkins: Yes it would be just `ros2 topic hz /path/to/topic`. If you only have RGB enabled then you are looking for ~10hz.

PR #3, opened 2023-03-30
Changes for ROS2

jbarry-bdai:
So far these are the changes I've identified to make the execution identical with the ROS2 repo.  They still need to be tested on robot.  List of tests to run can be found at https://github.com/bdaiinstitute/spot_ros2/pull/28.


heuristicus (approved):
Looks good, merging. This probably means the PR in spot_ros2 needs to update the submodule to point at main again.

jbarry-bdai:
(because we have an internal version of our repo that points to this branch by name i'm going to keep it around at least for a little while)

PR #5, opened 2023-04-04
Fix claim always returning true once lease object was initialised

heuristicus:
I was testing the claim function and it always returned true even if the lease was hijacked after we had acquired it. This should fix that problem.

The lease object's is_valid_lease property apparently is not talking about the lease being usable, it just means the proto is valid. I tried some other methods of checking if the lease was held by us. Checking the lease status held by the wallet is SELF_OWNER apparently doesn't work, because it's still true even when the lease is hijacked. So now I explicitly check the all-leases resource's lease owner contains spot_ros.

Instead of using the lease objects, I instead check the leases object of the robot and compare the owner of the `all-leases` lease to the name of the client.

This also fixes a missing underscore in the async idle `_continually_try_stand`.

I also made it so that the name of the client that acquires leases and so on is a constant, which may be useful since this appears in various places.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -367,7 +367,7 @@ class AsyncIdle(AsyncPeriodicQuery):
         # command can cause other commands to be interrupted before they get to start
         if (
             self._spot_wrapper.is_standing
-            and self._spot_wrapper.continually_try_stand
+            and self._spot_wrapper._continually_try_stand
             and not self._spot_wrapper.is_moving
             and self._spot_wrapper._last_trajectory_command is not None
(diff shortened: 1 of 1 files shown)


heuristicus:
On further testing, this still isn't functioning properly - the power on command asserts there is no lease for the body resource despite the lease list indicating that we have the lease.

jbarry-bdai (approved):
Oh no sorry about the typo!

jbarry-bdai:
We do use the claim functionality and it worked for us... we don't run the estop through the driver though.

peci1:
This PR broke auto_claim. This line in `def claim()` needs to check if `self.lease` is not None:

    for resource in self.lease:

PR #6, opened 2023-04-18
Draft: Refactoring code to be more modular

jeremysee2:
This continues work on the ROS1 version of the Spot wrapper to modularize the code and add testing to the wrapper, from https://github.com/heuristicus/spot_ros/issues/104 

List of breaking changes to the ROS interface:

1. front/side/rear images now follow the ROS2 implementation of synchronous calls instead of AsyncPeriodicQuery
2. All spot_arm calls have to be updated (e.g. `self.spot_wrapper.spot_arm.arm_stow()` instead of `self.spot_wrapper.arm_stow()`, adding the module name inbetween
3. All docking calls have to be updated
4. All GraphNav calls have to be updated
5. All Images calls have to be updated
6. The `list_world_objects` call has to be updated
7. `graph_nav_util` has been folded into the GraphNav class, with the BD copyright notice marking where the copied code starts

New features:
1. GraphNav has been significantly refactored into `navigate_init`, `navigate_to` and `navigate_route`
2. Spot Check has been added
3. 

Currently, the only unit tests are for some parts of graph_nav_util, written in `unittest` and triggered automatically via Github Actions CI.

The PR is not ready to merge, requires testing and comments.


jeremysee2:
Hand images use the old AsyncPeriodicQuery, because I don't want the ROS interface to be aware of whether the arm exists; this should be handled directly by the Wrapper.

heuristicus:
Thanks for the work you've put in on this. I'll take some time to review soon.

jeremysee2:
I've made PR's to prepare for the breaking changes, in both the ROS1 and ROS2 implementations of spot_ros:

ROS1: https://github.com/heuristicus/spot_ros/pull/126
ROS2: https://github.com/bdaiinstitute/spot_ros2/pull/41

jeremysee2:
Tested the code with the ROS1 and ROS2 wrapper, appears to work on the surface level (claim, stand, sit).

jeremysee2:
Unfortunately, I won't have access to the robot after 26 April, so I can't test it on hardware beyond that date.

spot_wrapper/spot_arm.py
```diff
@@ -0,0 +1,554 @@
...
+        """
+        Constructor for SpotArm class.
+        :param robot: Robot object
+        :param logger: Logger object
+        :param robot_params: Dictionary of robot parameters
+            - robot_params['is_standing']: True if robot is standing, False otherwise
+        :param robot_clients: Dictionary of robot clients
+            - robot_clients['robot_command_client']: RobotCommandClient object
+            - robot_clients['robot_command_method']: RobotCommand method
+        """
```
heuristicus, line 68:
```suggestion
        Constructor for SpotArm class.
        Args:
            robot: Robot object
            logger: Logger object
            robot_params: Dictionary of robot parameters
                          - robot_params['is_standing']: True if robot is standing, False otherwise
            robot_clients: Dictionary of robot clients
                           - robot_clients['robot_command_client']: RobotCommandClient object
                           - robot_clients['robot_command_method']: RobotCommand method
```

Google doc style rather than reST.

spot_wrapper/spot_config.py
heuristicus, line 1: Putting things into this file and then importing it with `from spot_config import *` might make it a little unclear where these things are coming from. I think we should prefer to be specific about what we're importing from the config.

If these things are only used in a single file it might make sense to have them there.

Practically speaking I think this is a cosmetic preference, so not sure it matters too much to have things here. We're using global variables anyway which is already not ideal, so putting them in once place makes sure they aren't scattered everywhere.

spot_wrapper/spot_graph_nav.py
```diff
@@ -0,0 +1,796 @@
...
+
+    def optmize_anchoring(self) -> typing.Tuple[bool, str]:
+        return self._optimize_anchoring()
+
+    ## Copied from spot-sdk/python/examples/graph_nav_command_line/graph_nav_command_line.py
+    # Copyright (c) 2020 Boston Dynamics, Inc.  All rights reserved.
+    #
+    # Downloading, reproducing, distributing or otherwise using the SDK Software
+    # is subject to the terms and conditions of the Boston Dynamics Software
+    # Development Kit License (20191101-BDSDK-SL).
```
heuristicus, line 147:
```suggestion
    ## Copied from https://github.com/boston-dynamics/spot-sdk/blob/master/python/examples/graph_nav_command_line/recording_command_line.py and https://github.com/boston-dynamics/spot-sdk/blob/master/python/examples/graph_nav_command_line/graph_nav_command_line.py with minor modifications
    # Copyright (c) 2020 Boston Dynamics, Inc.  All rights reserved.
    #
    # Downloading, reproducing, distributing or otherwise using the SDK Software
    # is subject to the terms and conditions of the Boston Dynamics Software
    # Development Kit License (20191101-BDSDK-SL).
```

Should make clear that we have made some modifications

spot_wrapper/spot_graph_nav.py
```diff
@@ -0,0 +1,796 @@
...
+                == map_processing_pb2.ProcessAnchoringResponse.STATUS_MAP_MODIFIED_DURING_PROCESSING
+            ):
+                return False, "Map modified during processing."
+            elif (
+                response.status
+                == map_processing_pb2.ProcessAnchoringResponse.STATUS_INVALID_GRAVITY_ALIGNMENT
+            ):
+                return False, "Invalid gravity alignment."
+            else:
+                return False, "Unknown error during anchoring optimization."
```
heuristicus, line 664:
```suggestion
            self._logger.error("Error optimizing {}".format(response))
            status_mapping = {
                map_processing_pb2.ProcessAnchoringResponse.STATUS_MISSING_WAYPOINT_SNAPSHOTS: "Missing waypoint snapshots.",
                map_processing_pb2.ProcessAnchoringResponse.STATUS_INVALID_GRAPH: "Invalid graph.",
                map_processing_pb2.ProcessAnchoringResponse.STATUS_OPTIMIZATION_FAILURE: "Optimization failure.",
                map_processing_pb2.ProcessAnchoringResponse.STATUS_INVALID_PARAMS: "Invalid parameters.",
                map_processing_pb2.ProcessAnchoringResponse.STATUS_CONSTRAINT_VIOLATION: "Constraint violation.",
                map_processing_pb2.ProcessAnchoringResponse.STATUS_MAX_ITERATIONS: "Max iterations, timeout.",
                map_processing_pb2.ProcessAnchoringResponse.STATUS_MAX_TIME: "Max time reached, timeout.",
                map_processing_pb2.ProcessAnchoringResponse.STATUS_INVALID_HINTS: "Invalid hints.",
                map_processing_pb2.ProcessAnchoringResponse.STATUS_MAP_MODIFIED_DURING_PROCESSING: "Map modified during processing.",
                map_processing_pb2.ProcessAnchoringResponse.STATUS_INVALID_GRAVITY_ALIGNMENT: "Invalid gravity alignment.",
            }
            if response.status in status_mapping:
                return False, status_mapping[response.status]

            return False, "Unknown error during anchoring optimization."
```

Another mapping to simplify if statements.

spot_wrapper/wrapper.py
```diff
@@ -518,21 +370,15 @@ def __init__(
         self._password = password
         self._hostname = hostname
         self._robot_name = robot_name
+        self._rates = rates
+        self._callbacks = callbacks
```
heuristicus, line 374:
```suggestion
        self._rates = rates or {}
        self._callbacks = callbacks or {}
```

Handle the case where rates and callbacks are unset in the init args (i.e. None).

spot_wrapper/wrapper.py
```diff
@@ -489,8 +341,8 @@ def __init__(
         logger: logging.Logger,
         start_estop: bool = True,
         estop_timeout: float = 9.0,
-        rates: typing.Optional[typing.Dict] = None,
-        callbacks: typing.Optional[typing.Dict] = None,
+        rates: typing.Dict[str, float] = {},
+        callbacks: typing.Dict[str, typing.Callable] = {},
```
heuristicus, line 345:
```suggestion
        rates: typing.Dict[str, float] = None,
        callbacks: typing.Dict[str, typing.Callable] = None,
```

Should avoid mutable default arguments, see https://stackoverflow.com/questions/1132941/least-astonishment-and-the-mutable-default-argument. It probably doesn't matter since the wrapper is only initialised a single time, but better to avoid. This does mean that technically the type hint is incorrect.

spot_wrapper/wrapper.py
```diff
@@ -648,255 +439,298 @@ def __init__(
...
+                    "power_client": self._power_client,
+                    "lease_client": self._lease_client,
+                    "image_client": self._image_client,
+                    "estop_client": self._estop_client,
+                    "docking_client": self._docking_client,
+                    "spot_check_client": self._spot_check_client,
+                    "robot_command_method": self._robot_command,
+                    "world_objects_client": self._world_objects_client,
+                    "manipulation_client": self._manipulation_client,
+                }
```
heuristicus, line 505: This makes me think that we should implement something like a client manager which can be used to access these things, and probably initialise them as well. But that's probably for later rather than in this PR.

spot_wrapper/wrapper.py
```diff
@@ -648,255 +439,298 @@ def __init__(
...
+            "is_standing": self._is_standing,
+            "is_sitting": self._is_sitting,
+            "is_moving": self._is_moving,
+            "at_goal": self._at_goal,
+            "near_goal": self._near_goal,
+            "robot_id": self._robot_id,
+            "estop_timeout": self._estop_timeout,
+            "rates": self._rates,
+            "callbacks": self._callbacks,
+        }
```
heuristicus, line 571: This should probably also be an object rather than a dict, again something to consider later.

heuristicus (commented):
Some minor changes I've added, but other than that I think at least visually it looks good. Still need to test with our robot.

spot_wrapper/spot_check.py
```diff
@@ -0,0 +1,260 @@
...
+            resp.error == spot_check_pb2.SpotCheckFeedbackResponse.ERROR_REVERT_FAILURE
+        ):
+            return False, "Robot failed to revert calibration"
+        elif resp.error == spot_check_pb2.SpotCheckFeedbackResponse.ERROR_FGKC_FAILURE:
+            return False, "Robot failed to do flat ground kinematic calibration"
+        elif (
+            resp.error
+            == spot_check_pb2.SpotCheckFeedbackResponse.ERROR_GRIPPER_CAL_TIMEOUT
+        ):
+            return False, "Timeout during gripper calibration"
```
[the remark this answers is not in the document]
jeremysee2: Agreed, done in [392da89](https://github.com/bdaiinstitute/spot_wrapper/pull/6/commits/392da898d027fc716a17f1cd7ff487183cc192cb)

spot_wrapper/spot_arm.py
jeremysee2: Agreed, done in [392da89](https://github.com/bdaiinstitute/spot_wrapper/pull/6/commits/392da898d027fc716a17f1cd7ff487183cc192cb)

spot_wrapper/spot_graph_nav.py
jeremysee2: Agreed, done in [392da89](https://github.com/bdaiinstitute/spot_wrapper/pull/6/commits/392da898d027fc716a17f1cd7ff487183cc192cb)

spot_wrapper/spot_graph_nav.py
jeremysee2: Agreed, done in [392da89](https://github.com/bdaiinstitute/spot_wrapper/pull/6/commits/392da898d027fc716a17f1cd7ff487183cc192cb)

spot_wrapper/wrapper.py
jeremysee2: Agreed, done in [392da89](https://github.com/bdaiinstitute/spot_wrapper/pull/6/commits/392da898d027fc716a17f1cd7ff487183cc192cb)

spot_wrapper/wrapper.py
jeremysee2: Agreed, done in [392da89](https://github.com/bdaiinstitute/spot_wrapper/pull/6/commits/392da898d027fc716a17f1cd7ff487183cc192cb)

spot_wrapper/wrapper.py
jeremysee2: Agreed, this works for now (since almost everything is an object in Python), but it would be nice to be explicit about what needs to be defined at runtime before going to operation.

spot_wrapper/wrapper.py
jeremysee2: Agreed, this works for now (since almost everything is an object in Python), but it would be nice to be explicit about what states are shared at runtime before going to operation.

spot_wrapper/spot_config.py
jeremysee2: Checked that all globals are actually used only within one file at a time, so moved them to the corresponding source files in [22d1084](https://github.com/bdaiinstitute/spot_wrapper/pull/6/commits/22d108496e0906e798e1324fabe113378cd8e806).

jeremysee2:
Apart from testing on hardware, the main concern is to synchronise the hand image publishing with https://github.com/bdaiinstitute/spot_wrapper/pull/7

This PR uses the old style (async) where the wrapper checks if the arm exists, then starts publishing. The other PR relies on the ROS wrapper to check if the arm exists, then trigger image requests.

heuristicus:
@DavidWatkins @jbarry-bdai I'd appreciate your input on this one. I think it would be to our benefit generally to make the code more modular, since changes we make would be localised to classes which are associated with specific functions. I know @jeremysee2 has put a lot of work into this and I wouldn't want it to be overlooked.

spot_wrapper/spot_arm.py
```diff
@@ -0,0 +1,561 @@
...
+    Attributes:
+        client: The Client to a service on the robot
+        logger: Logger object
+        rate: Rate (Hz) to trigger the query
+        callback: Callback function to call when the results of the query are available
+    """
+
+    def __init__(self, client, logger, rate, callback, image_requests):
+        super(AsyncImageService, self).__init__(
+            "robot_image_service", client, logger, period_sec=1.0 / max(rate, 1.0)
```
davidwatkins-bdai, line 43: Add typing

spot_wrapper/spot_arm.py
```diff
@@ -0,0 +1,561 @@
...
+        # Joint3: 0.0 arm straight. moves the arm down
+        # RANGE: 0.0 -> 3.1415
+        # Joint4: 0.0 middle position. negative: moves ccw, positive moves cw
+        # RANGE: -2.79253 -> 2.79253
+        # # Joint5: 0.0 gripper points to the front. positive moves the gripper down
+        # RANGE: -1.8326 -> 1.8326
+        # Joint6: 0.0 Gripper is not rolled, positive is ccw
+        # RANGE: -2.87 -> 2.87
+        # Values after unstow are: [0.0, -0.9, 1.8, 0.0, -0.9, 0.0]
+        if abs(joint_targets[0]) > 3.14:
```
davidwatkins-bdai, line 238: All of these values should be loaded from a configuration file or the URDF itself.

spot_wrapper/spot_arm.py
```diff
@@ -0,0 +1,561 @@
...
+
+    def force_trajectory(self, data) -> typing.Tuple[bool, str]:
+        try:
+            success, msg = self.ensure_arm_power_and_stand()
+            if not success:
+                self._logger.info(msg)
+                return False, msg
+            else:
+
+                def create_wrench_from_msg(forces, torques):
```
davidwatkins-bdai, line 311: Not clear why this function is defined here

spot_wrapper/spot_arm.py
```diff
@@ -0,0 +1,561 @@
...
+
+        except Exception as e:
+            return False, "Exception occured while gripper was moving"
+
+        return True, "Closed gripper successfully"
+
+    def gripper_angle_open(self, gripper_ang: float) -> typing.Tuple[bool, str]:
+        # takes an angle between 0 (closed) and 90 (fully opened) and opens the
+        # gripper at this angle
+        if gripper_ang > 90 or gripper_ang < 0:
```
davidwatkins-bdai, line 415: These values should be loaded from a URDF

spot_wrapper/spot_arm.py
```diff
@@ -0,0 +1,561 @@
...
+                    synchronized_command=synchronized_command
+                )
+
+                command = RobotCommandBuilder.build_synchro_command(robot_command)
+
+                # Send the request
+                self._robot_command_client.robot_command(robot_command)
+                self._logger.info("Moving arm to position.")
+
+                time.sleep(2.0)
```
davidwatkins-bdai, line 500: Why is there a sleep in this function?

spot_wrapper/spot_graph_nav.py
```diff
@@ -0,0 +1,759 @@
...
+            }
+            if response.status in status_mapping:
+                return False, status_mapping[response.status]
+            return False, "Unknown error during anchoring optimization."
+
+    def _id_to_short_code(self, id: str):
+        """Convert a unique id to a 2 letter short code."""
+        tokens = id.split("-")
+        if len(tokens) > 2:
+            return "%c%c" % (tokens[0][0], tokens[1][0])
```
davidwatkins-bdai, line 633: Use f-strings instead

davidwatkins-bdai (commented):
I have made some comments on cleaning up the code as well as some design decisions. These are extensive changes that would require us to test this against our existing stack. Looking at the corresponding change to the `spot_ros2` driver shows that this shouldn't be an incredible lift to integrate. I will try to get these changes tested.

spot_wrapper/spot_arm.py
heuristicus: I can't be sure and I've not used the arm at all, but perhaps it's necessary to make sure the arm is actually in position before the command returns? It's not a good way to do this though. Perhaps there is a way to check the arm command feedback. The status message at https://dev.bostondynamics.com/protos/bosdyn/api/proto_reference#robotcommandfeedbackstatus-status might be a mechanism? I did a similar thing with trajectory commands.

jeremysee2:
Most of the code was directly transferred from the old spot_wrapper code, so there isn't a strong reason behind many of the decision choices highlighted above. I don't have access to the robot anymore, so I won't be able to help with testing the code on hardware.

heuristicus:
@jeremysee2 do you have time to integrate the changes which have been made to the wrapper since your last updates? I've made a start on it but wanted to make sure we're not duplicating effort.

jeremysee2:
I'll look back at the commits merged since I started working on it to make sure those features are retained, then I'll notify for testing.

spot_wrapper/spot_arm.py
amessing-bdai: There  is a [`block_until_arm_arrives`](https://github.com/boston-dynamics/spot-sdk/blob/d1d743bbe419fddecea6c58865a681f19080395a/python/bosdyn-client/src/bosdyn/client/robot_command.py#L1731) function in the spot-sdk, which just polls the status messages. I think it would be better to use this function or write our own version.

heuristicus: Good find, we should use the SDK functionality rather than implementing our own.

heuristicus:
@jeremysee2 in the interest of getting this done quickly I'm going to work on it today and try and get the changes to the wrapper merged in to this branch.

@kaiyu-zheng This PR is not intended to override the new features.

jeremysee2:
@heuristicus thank you, just to re-affirm that this doesn't override new features, but refactor them.

kaiyu-zheng:
@jeremysee2  Could you please provide some background on the circumstances that led to these changes being made?  

Given that you're no longer using the robot, any issues stemming from this PR will fall on us to resolve. I simply want to emphasize that this PR contains substantial changes with minimal testing, which is a cause for concern. Please understand if I seem a bit resistant, it's merely out of caution.

jeremysee2:
I understand the need for caution, here's the background behind the PR:

https://github.com/heuristicus/spot_ros/issues/104 

A few months ago, I contacted @heuristicus and offered to refactor the codebase to make it more modular and easier to contribute to. I was working with Spot as part of my Masters' thesis, so the only testing I did was using it myself.

This was something that was brought up late last year: https://github.com/heuristicus/spot_ros/issues/90

heuristicus:
From my perspective, I'd been concerned about the monolithic wrapper for some time and hoped to modify it myself after working on implementing the spot_cam wrapper, which I built from the start with the intent of it being as modular as possible. After @jeremysee2 offered to do it I was happy to hand it over to him since it was something I didn't really have much time to focus on.

Perhaps it would be useful for @jeremysee2 to clarify what kind of testing he did up to this point. Once I have finished merging the most recent wrapper changes I will test some components on my side, but I don't have access to a robot with an arm and don't use graphnav at all so can't test those parts.

This is a significant change in terms of structure, but I think it's an important step to making it easier to maintain and understand.

jeremysee2:
In terms of testing, I've tested all the ROS services from command line, with a video demonstrating some of them: https://youtu.be/51ZHHd4vcAA

As part of a wider application, I demonstrated that the camera and LIDAR data can be used with RTAB-Map and the ROS navigation stack: https://youtu.be/gDW2GsG0zOw

heuristicus:
I've now completed the preliminary merge of all changes to the wrapper. I'm using the spot_ros repo and its corresponding PR https://github.com/heuristicus/spot_ros/pull/126 for testing. I have tested lease claim/release, powering on and off, gentle and hard stops and releasing the hard stop, sit and stand commands. Those all seem to work without issue, but are also the most simple.

The relevant PR for ros2 is https://github.com/bdaiinstitute/spot_ros2/pull/41.

Here's a list of what I think needs testing:

- [ ] Simple commands (lease, power, sit/stand)
- [ ] Trajectory
- [ ] Graph nav
- [ ] Docking
- [ ] Manipulation
- [ ] Spot check

Unfortunately I don't have access to a spot with docking or manipulation, and I have never used graph nav so likely can't test that either.

jbarry-bdai:
@kaiyu-zheng can you check graph nav with ROS2 with this PR when you get a chance?  If you can also run the HIL test that would be amazing but I can do it when I get back too :)

jbarry-bdai (commented):
I did some spot checking (haha).  I agree that I think the most important thing here will be to run the testing.  At a minimum we need to do the normal two BDAII HIL tests.  We should probably also run the examples from spot ros2 (although I think I broke those when I introduced ros utilities so need to go back and fix that first).

I can try to do this when I'm back on site next week or the week after unless someone gets to it first.

spot_wrapper/wrapper.py
```diff
@@ -722,234 +466,228 @@ def __init__(
...
 
-            self._idle_task = AsyncIdle(
-                self._robot_command_client, self._logger, 10.0, self
-            )
-            self._estop_monitor = AsyncEStopMonitor(
-                self._estop_client, self._logger, 20.0, self
-            )
-            self._world_objects_task = AsyncWorldObjects(
-                self._world_objects_client,
+        if self._robot.has_arm():
```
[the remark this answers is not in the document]
myeatman-bdai: I think so, but I'm leaving another comment on the property method for a custom exception class for trying to access the spot arm if one does not exist.

spot_wrapper/wrapper.py
```diff
@@ -990,104 +728,127 @@ def authenticate(robot, username, password, logger):
...
-    def logger(self):
+    def spot_images(self) -> SpotImages:
+        """Return SpotImages instance"""
+        return self._spot_images
+
+    @property
+    def spot_arm(self) -> SpotArm:
+        """Return SpotArm instance"""
+        if not self._robot.has_arm():
+            raise Exception("SpotArm is not available on this robot")
```
myeatman-bdai, line 747: Please make a custom exception class called "MissingSpotArm" or the like.

myeatman-bdai (changes requested):
Please see my comment asking for a custom exception class.

kaiyu-zheng:
> Ran the BDAII tests with that and the associated HIL tests
> 
> * The bike repair test succeeded
> * Graph Nav test failed with the below error @kaiyu-zheng
> 
> ```
> [spot_ros2-1]   File "/workspaces/bdai/ws/build/spot_wrapper/spot_wrapper/spot_graph_nav.py", line 544, in clear_graph                                                                                             
> [spot_ros2-1]     result = self._graph_nav_client.clear_graph(lease=self._lease.lease_proto)                                                                                                                       
> [spot_ros2-1] AttributeError: 'SpotGraphNav' object has no attribute '_lease'                                                                                                                                      
> ```
Looks like SpotWrapper originally has a `self._lease` field, and it has functions like `getLease` that set that field. How is this handled now?

spot_wrapper/wrapper.py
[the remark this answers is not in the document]
amessing-bdai: The issue wasn't on this side. It was client-side checking the parameter server

amessing-bdai:
Also, I want to reiterate that all of the arm functions should use `block_until_arm_arrives` instead of `time.sleep`

jeremysee2:
> > Ran the BDAII tests with that and the associated HIL tests
> > 
> > * The bike repair test succeeded
> > * Graph Nav test failed with the below error @kaiyu-zheng
> > 
> > ```
> > [spot_ros2-1]   File "/workspaces/bdai/ws/build/spot_wrapper/spot_wrapper/spot_graph_nav.py", line 544, in clear_graph                                                                                             
> > [spot_ros2-1]     result = self._graph_nav_client.clear_graph(lease=self._lease.lease_proto)                                                                                                                       
> > [spot_ros2-1] AttributeError: 'SpotGraphNav' object has no attribute '_lease'                                                                                                                                      
> > ```
> 
> Looks like SpotWrapper originally has a `self._lease` field, and it has functions like `getLease` that set that field. How is this handled now?

Currently, the `self._lease` field is created when initializing GraphNav with a map, or when telling the robot to navigate to a specific waypoint. Clearing the graph (currently) does not attempt to claim the lease.

It's not clear if `clear_graph` [requires the lease](https://dev.bostondynamics.com/python/bosdyn-client/src/bosdyn/client/graph_nav#bosdyn.client.graph_nav.GraphNavClient.clear_graph).

If clearing the graph doesn't require claiming the lease, I recommend wrapping that call in an `if` statement to avoid this problem. If it does, we can add in the following snippet to claim the lease:

```
self._lease = self._lease_wallet.get_lease()
```

spot_wrapper/wrapper.py
jeremysee2: Added in 0ec85565f653680fdcdf1dcfd65dd93803faedda

jeremysee2:
> 651fed6

Addressed with the addition of `self._get_lease` in the clear_graph method [651fed6](https://github.com/bdaiinstitute/spot_wrapper/pull/6/commits/651fed617ef016c3a33e007b8097a4e5d07e8232)

jeremysee2:
> > > Ran the BDAII tests with that and the associated HIL tests
> > > 
> > > * The bike repair test succeeded
> > > * Graph Nav test failed with the below error @kaiyu-zheng
> > > 
> > > ```
> > > [spot_ros2-1]   File "/workspaces/bdai/ws/build/spot_wrapper/spot_wrapper/spot_graph_nav.py", line 544, in clear_graph                                                                                             
> > > [spot_ros2-1]     result = self._graph_nav_client.clear_graph(lease=self._lease.lease_proto)                                                                                                                       
> > > [spot_ros2-1] AttributeError: 'SpotGraphNav' object has no attribute '_lease'                                                                                                                                      
> > > ```
> > 
> > 
> > Looks like SpotWrapper originally has a `self._lease` field, and it has functions like `getLease` that set that field. How is this handled now?
> 
> Currently, the `self._lease` field is created when initializing GraphNav with a map, or when telling the robot to navigate to a specific waypoint. Clearing the graph (currently) does not attempt to claim the lease.
> 
> It's not clear if `clear_graph` [requires the lease](https://dev.bostondynamics.com/python/bosdyn-client/src/bosdyn/client/graph_nav#bosdyn.client.graph_nav.GraphNavClient.clear_graph).
> 
> If clearing the graph doesn't require claiming the lease, I recommend wrapping that call in an `if` statement to avoid this problem. If it does, we can add in the following snippet to claim the lease:
> 
> ```
> self._lease = self._lease_wallet.get_lease()
> ```

Addressed with the addition of self._get_lease in the clear_graph method [651fed6](https://github.com/bdaiinstitute/spot_wrapper/pull/6/commits/651fed617ef016c3a33e007b8097a4e5d07e8232)

heuristicus:
> Also, I want to reiterate that all of the arm functions should use `block_until_arm_arrives` instead of `time.sleep`

Addressed this in [e6df481](https://github.com/bdaiinstitute/spot_wrapper/pull/6/commits/e6df4814f459baebc3d7453bac98b61164f49949). The grasp_3d command comes from the manipulation API and doesn't seem to have an equivalent function.

spot_wrapper/wrapper.py
```diff
@@ -722,234 +466,228 @@ def __init__(
...
-            while not initialised:
+        if not self._robot:
+            self._logger.error("Failed to create robot object")
+            self._valid = False
+            return
+
+        # Clients
+        self._logger.info("Creating clients...")
+        initialised = False
+        while not initialised:
```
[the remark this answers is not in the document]
heuristicus: I'm not sure I follow. I used a while loop here as part of the startup procedure to make it so that it's not necessary to wait for the robot to be started up fully. If it isn't started, some of these services will not be constructed and thus cause exceptions, which we want to ignore because we know that a good startup should not result in any exceptions. The while loop exits when everything initialises correctly.

Unless you mean that we should limit the number of attempts?

kaiyu-zheng:
> > > Ran the BDAII tests with that and the associated HIL tests
> > > 
> > > * The bike repair test succeeded
> > > * Graph Nav test failed with the below error @kaiyu-zheng
> > > 
> > > ```
> > > [spot_ros2-1]   File "/workspaces/bdai/ws/build/spot_wrapper/spot_wrapper/spot_graph_nav.py", line 544, in clear_graph                                                                                             
> > > [spot_ros2-1]     result = self._graph_nav_client.clear_graph(lease=self._lease.lease_proto)                                                                                                                       
> > > [spot_ros2-1] AttributeError: 'SpotGraphNav' object has no attribute '_lease'                                                                                                                                      
> > > ```
> > 
> > 
> > Looks like SpotWrapper originally has a `self._lease` field, and it has functions like `getLease` that set that field. How is this handled now?
> 
> Currently, the `self._lease` field is created when initializing GraphNav with a map, or when telling the robot to navigate to a specific waypoint. Clearing the graph (currently) does not attempt to claim the lease.
> 
> It's not clear if `clear_graph` [requires the lease](https://dev.bostondynamics.com/python/bosdyn-client/src/bosdyn/client/graph_nav#bosdyn.client.graph_nav.GraphNavClient.clear_graph).
> 
> If clearing the graph doesn't require claiming the lease, I recommend wrapping that call in an `if` statement to avoid this problem. If it does, we can add in the following snippet to claim the lease:
> 
> ```
> self._lease = self._lease_wallet.get_lease()
> ```

Since only one node can possess the lease to a Spot at a time, would it be better if the lease wollet is owned by the SpotWrapper object, instead of by the individual components?  Correct me if I am misunderstanding something.

heuristicus:
> > > > Ran the BDAII tests with that and the associated HIL tests
> > > > 
> > > > * The bike repair test succeeded
> > > > * Graph Nav test failed with the below error @kaiyu-zheng
> > > > 
> > > > ```
> > > > [spot_ros2-1]   File "/workspaces/bdai/ws/build/spot_wrapper/spot_wrapper/spot_graph_nav.py", line 544, in clear_graph                                                                                             
> > > > [spot_ros2-1]     result = self._graph_nav_client.clear_graph(lease=self._lease.lease_proto)                                                                                                                       
> > > > [spot_ros2-1] AttributeError: 'SpotGraphNav' object has no attribute '_lease'                                                                                                                                      
> > > > ```
> > > 
> > > 
> > > Looks like SpotWrapper originally has a `self._lease` field, and it has functions like `getLease` that set that field. How is this handled now?
> > 
> > 
> > Currently, the `self._lease` field is created when initializing GraphNav with a map, or when telling the robot to navigate to a specific waypoint. Clearing the graph (currently) does not attempt to claim the lease.
> > It's not clear if `clear_graph` [requires the lease](https://dev.bostondynamics.com/python/bosdyn-client/src/bosdyn/client/graph_nav#bosdyn.client.graph_nav.GraphNavClient.clear_graph).
> > If clearing the graph doesn't require claiming the lease, I recommend wrapping that call in an `if` statement to avoid this problem. If it does, we can add in the following snippet to claim the lease:
> > ```
> > self._lease = self._lease_wallet.get_lease()
> > ```
> 
> Since only one node can possess the lease to a Spot at a time, would it be better if the lease wallet is owned by the SpotWrapper object, instead of by the individual components? Correct me if I am misunderstanding something.

I think that because it's part of the same node, getting a lease object should just return the same values. But I'm not certain. I tried doing some stuff with the lease wallet previously and it didn't behave the way I expected, so there may be some complications.

amessing-bdai:
> > Also, I want to reiterate that all of the arm functions should use `block_until_arm_arrives` instead of `time.sleep`
> 
> Addressed this in [e6df481](https://github.com/bdaiinstitute/spot_wrapper/pull/6/commits/e6df4814f459baebc3d7453bac98b61164f49949). The grasp_3d command comes from the manipulation API and doesn't seem to have an equivalent function.

That's fine. You are basically doing the same thing that they do internally. Also, `spot_ros2` doesn't use the `grasp_3d` command and instead uses a more general `manipulation_api` command. I might still recommend that you have a timeout option similar to the [existing block](https://github.com/boston-dynamics/spot-sdk/blob/7b0a06c3ae6e23aca03aad85af883761a257c8b1/python/bosdyn-client/src/bosdyn/client/robot_command.py#L1911) commands.

amessing-bdai:
> > Since only one node can possess the lease to a Spot at a time, would it be better if the lease wallet is owned by the SpotWrapper object, instead of by the individual components? Correct me if I am misunderstanding something.
> 
> I think that because it's part of the same node, getting a lease object should just return the same values. But I'm not certain. I tried doing some stuff with the lease wallet previously and it didn't behave the way I expected, so there may be some complications.

I don't thing whether it is part of the same node matters so much. I think because we created the lease client from a single `robot` instances, all leases acquired/taken have the same owner. I think it will be the same lease, but possibly have an updated request time. The only time it would be a new lease would be if something else took the lease (e.g. the tablet or another program) and then you took it back. 

I'm not sure there is a different between using the get_lease or just leaving the default value of none which says it will use the client's lease. When the various functions create subleases and pass those to the graph_nav_client is when I think having the lease_wallet matters. 

Also, I am also curious about having the lease_keepalive internally and if there are any conflicts with the lease_keepalive that is held in SpotWrapper. Potentially there are times when there are two threads that are trying to keep the lease alive. Probably doesn't hurt in terms of messing with the lease, but just potentially extra computation and communication. Also, there are times where various functions shutdown the internal keepalive, usually just before they pass a sublease to the graph nav client, so I'd be curious if the external keepalive (which presumably is still running) creates a conflict.

spot_wrapper/wrapper.py
```diff
@@ -729,252 +474,228 @@ def __init__(
...
+        )
+        self._estop_monitor = AsyncEStopMonitor(
+            self._estop_client, self._logger, 20.0, self
+        )
 
-            # Store the most recent knowledge of the state of the robot based on rpc calls.
-            self._init_current_graph_nav_state()
+        self._estop_endpoint = None
+        self._estop_keepalive = None
+        self._robot_id = None
```
amessing-bdai, line 619: The property down below uses the robot_params dict instead of self._robot_id? If this isn't used anywhere they it should be removed to avoid confusion

heuristicus: I've made it so that the wrapper class no longer uses the convenience dict for accessing these values or returning them from properties. It still has the dict to pass to things which it instantiates.

spot_wrapper/spot_arm.py
heuristicus: I don't have any experience using the arm, but are these joint limit checks even necessary? Based on other commands, the motion should be limited on the robot anyway. 

If we do need them, I think the way to access is to use [`get_robot_hardware_configuration`](https://dev.bostondynamics.com/python/bosdyn-client/src/bosdyn/client/robot_state.html#bosdyn.client.robot_state.RobotStateClient.get_robot_hardware_configuration) to retrieve the robot skeleton, which returns [`RobotHardwareConfigurationResponse`](https://dev.bostondynamics.com/protos/bosdyn/api/proto_reference.html?highlight=urdf#robothardwareconfigurationresponse). Probably the best way to access the URDF is to use `urdfpy`, which seems to allow easy access to joint limits (actually it seems that library is no longer maintained, perhaps https://github.com/fishbotics/urchin/ would be better to use). It can only load URDF from file, so would need to save to a tmp file first

heuristicus:
I've fixed up some of the issues which were mentioned by @amessing-bdai and @myeatman-bdai. I'd like to get this merged and it would be good to know if there are any other concerns which would be considered blocking so I can try and fix them.

heuristicus:
Per the discussion on slack, I will close this PR and split it out into individual PRs for each module which achieves the same result but is easier to test.

PR #11, opened 2023-05-08
[SW-127] Add function to get images by cameras

kaiyu-zheng:
**Motivation:** Currently, SpotWrapper only gets images from all cameras. This impacts performance especially when the images are expensive to compute (e.g. registered depth images). It is a basic and useful feature to allow a user to specify from which camera(s) and for which image types they want the images.

**What's New:**  This PR adds an additional function to SpotWrapper, `get_images_by_cameras`.  This allows a user to:
* Specify which cameras they want to get images from (e.g. 'frontleft' and 'back')
   ```python
   spot_wrapper.get_images_by_cameras(['frontleft', 'back'])
   ``` 
   (this will get images of types RGB, depth, and depth_registered for frontleft and back cameras)

* Specify which cameras they want to get images from and what image types (e.g. 'visual' and 'depth_registered' for 'frontleft', 'visual' for 'back'):
   ```python
   spot_wrapper.get_images_by_cameras([('frontleft', ['visual', 'depth_registered']),
                                       ('back', ['visual'])])
   ```
This supports, for example, spot_ros2 to add a new parameter `cameras_used` which allows a user to select which cameras to get & publish images for.  

This PR goes together with https://github.com/bdaiinstitute/spot_ros2/pull/52

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -2533,18 +2533,26 @@ class SpotWrapper:
     ) -> typing.Optional[typing.Union[ImageBundle, ImageWithHandBundle]]:
         return self.get_images(self._depth_registered_image_requests)
 
-    def get_images_by_cameras(self, camera_sources):
+    def get_images_by_cameras(
+            self, camera_sources: typing.List[CameraSource]
+    ) -> typing.List[ImageEntry]:
         """Calls the GetImage RPC using the image client with requests
         corresponding to the given cameras.
 
-        Args: Accepts either a list of strings (e.g., ['frontleft', 'hand'])
-                or a list of tuples (e.g. [('frontleft', ['visual', 'depth'], ...]
+        Args:
+           camera_sources: a list of CameraSource objects. There are two
+               possibilities for each item in this list. Either it is
+               CameraSource(camera=Camera.front) or
+               CameraSource(camera=Camera.front, image_types=[ImageType.visual, ImageType.depth_registered)
+
                 - If the former is provided, the image requests will include all
                   image types for each specified camera.
                 - If the latter is provided, the image requests will be
                   limited to the specified image types for each corresponding
-                  camera.  Valid image types are 'visual', 'depth', 'depth_registered'
+                  camera.
+
               Note that duplicates of camera names are not allowed.
+
         Returns:
             a list, where each entry is (camera_name, image_type, image_response)
                 e.g. ('frontleft', 'visual', image_response)
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -119,79 +119,53 @@ ImageWithHandBundle = namedtuple(
     "ImageBundle", ["frontleft", "frontright", "left", "right", "back", "hand"]
 )
 
+IMAGE_SOURCES_BY_CAMERA = {
+    "frontleft": {
+        "visual": "frontleft_fisheye_image",
+        "depth": "frontleft_depth",
+        "depth_registered": "frontleft_depth_in_visual_frame"
+    },
+    "frontright": {
+        "visual": "frontright_fisheye_image",
+        "depth": "frontright_depth",
+        "depth_registered": "frontright_depth_in_visual_frame"
+    },
+    "left": {
+        "visual": "left_fisheye_image",
+        "depth": "left_depth",
+        "depth_registered": "left_depth_in_visual_frame"
+    },
+    "right": {
+        "visual": "right_fisheye_image",
+        "depth": "right_depth",
+        "depth_registered": "right_depth_in_visual_frame"
+    },
+    "back": {
+        "visual": "back_fisheye_image",
+        "depth": "back_depth",
+        "depth_registered": "back_depth_in_visual_frame"
+    },
+    "hand": {
+        "visual": "hand_color_image",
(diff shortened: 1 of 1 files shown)


spot_wrapper/wrapper.py
```diff
@@ -2437,3 +2498,70 @@ def get_depth_registered_images(
         self,
     ) -> typing.Optional[typing.Union[ImageBundle, ImageWithHandBundle]]:
         return self.get_images(self._depth_registered_image_requests)
+
+    def get_images_by_cameras(self, camera_sources):
```
amessing-bdai, line 2502: type hinting?

spot_wrapper/wrapper.py
```diff
@@ -2437,3 +2498,70 @@ def get_depth_registered_images(
...
+              Note that duplicates of camera names are not allowed.
+        Returns:
+            a list, where each entry is (camera_name, image_type, image_response)
+                e.g. ('frontleft', 'visual', image_response)
+        """
+        # Build image requests
+        image_requests = []
+        source_types = []
+        for item in camera_sources:
+            if type(item) == str:
```
amessing-bdai, line 2522: Is this better than isinstance?

amessing-bdai (approved):
Looks good. Please add the type hinting and then we can merge

spot_wrapper/wrapper.py
```diff
@@ -117,6 +117,42 @@
     "ImageBundle", ["frontleft", "frontright", "left", "right", "back", "hand"]
 )
 
+# Added to support getting images for given cameras
+IMAGE_SOURCES_BY_CAMERA = {
```
davidwatkins-bdai, line 121: All of these should be an enum rather than strings.

spot_wrapper/wrapper.py
davidwatkins-bdai: Agreed, please use isinstance

spot_wrapper/wrapper.py
```diff
@@ -2437,3 +2498,70 @@ def get_depth_registered_images(
...
+            for image_type in image_types:
+                image_requests.append(
+                    self._image_requests_by_camera[camera_name][image_type]
+                )
+                source_types.append((camera_name, image_type))
+
+        # Send image requests
+        try:
+            image_responses = self._image_client.get_image(image_requests)
+        except UnsupportedPixelFormatRequestedError as e:
```
davidwatkins-bdai, line 2550: Delete "as e" if you aren't using e

spot_wrapper/wrapper.py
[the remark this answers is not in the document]
davidwatkins-bdai: camera_sources would then be a list[CameraSource] type and you can do significantly fewer checks down below.

davidwatkins-bdai (approved):
Add type hints and replace camera source with an enum.

spot_wrapper/wrapper.py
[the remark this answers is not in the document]
kaiyu-zheng: see comment above.

spot_wrapper/wrapper.py
kaiyu-zheng: While typing the code below, I think.... typo can still happen.. (unless, we do static checking of course).
```python
IMAGE_SOURCES_BY_CAMERA = {
    Camera.frontleft: {
        ImageType.visual: ImageSource.frontleft_fisheye_image,
        ImageType.depth: ImageSource.frontleft_depth,
        ImageType.depth_registered: frontleft_depth_in_visual_frame,
    },
   ...
```

kaiyu-zheng: In any case, I did this. See the latest commit

ahoelscher-bdai: I don't think we should use an enum here. The underlying spot API provides strings, and is actually somewhat flexible with them. For example, you can define image sources for extensions that you attach to spot. If we do that, then we'd need to add a new enum here.

ahoelscher-bdai: I would spend the time to extend the spot_wrapper to provide the list of camera sources, and use that as the list. If you don't want to spend that time, I would open a jira ticket, and hard-code a list of camera sources at the top, with a comment linking to that jira ticket.

kaiyu-zheng: I was previously hard-coding the strings to be consistent with the other code in spot wrapper (because there's already a lot of hard-coded source names).  
I would be fine with leaving hard-coded strings at the top of spot wrapper because those strings would be written only once.   

That said, I'll remove the Enum for ImageSource and revert back to the strings. Those strings shouldn't need to be typed by the user of SpotWrapper anyways. I could open a JIRA ticket in the PickNik board about this.

I think the dataclass `CameraSource` is worthwhile to add, which clarifies the input type to `get_images_by_cameras`.

ahoelscher-bdai: Sounds good to me!

kaiyu-zheng: Ticket created: https://theaiinstitute.atlassian.net/browse/PN-5

I will mark this as resolved.  I reverted adding the enums, and tested again on hardware.  The functionality works.

amessing-bdai (approved):

PR #12, opened 2023-05-11
[SW-141] Checking edge cases in upload_graph

kaiyu-zheng:
This PR is for part of SW-138. Adds a service for upload graph, given a file path. This PR goes together with one in bdai and one in spot_wrapper. The test is in bdai.

When the service handler is called, the following driver log happens:
```
[spot_ros2-1] [INFO] [1683815953.577940734] [Gosu.spot_ros2]: Uploading GraphNav map: /workspaces/bdai/ws/src/spot_maps/maps/bdai5picasso
[spot_ros2-1] [INFO] [1683815953.579430977] [spot_wrapper]: Loading the graph from disk into local storage...
[spot_ros2-1] [INFO] [1683815953.580826267] [spot_wrapper]: Loaded graph has 23 waypoints and 26 edges
[spot_ros2-1] [INFO] [1683815953.587187860] [spot_wrapper]: Uploading the graph and snapshots to the robot...
[spot_ros2-1] [INFO] [1683815953.877267127] [spot_wrapper]: Uploaded snapshot_edgy-egg-xQaSq54VuWCgFsWe5ByY9g==
[spot_ros2-1] [INFO] [1683815954.112413740] [spot_wrapper]: Uploaded snapshot_fuggy-kelp-oOq99bWbMLTEWBJRRv7xPw==
...
[spot_ros2-1] [INFO] [1683815958.734839673] [Gosu.spot_ros2]: Uploaded
```

For context, "upload graph" corresponds to a RPC call [in Spot SDK](https://dev.bostondynamics.com/_modules/bosdyn/client/graph_nav#GraphNavClient.upload_graph:~:text=%5Bdocs%5D-,def%20upload_graph,-(self%2C)) that uploads a map file to the GraphNav server running on Spot.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -1,6 +1,7 @@
 import functools
 import logging
 import math
+import os
 import time
 import traceback
 import typing
@@ -812,11 +813,7 @@ class SpotWrapper:
                     )
 
             # Store the most recent knowledge of the state of the robot based on rpc calls.
-            self._current_graph = None
-            self._current_edges = dict()  # maps to_waypoint to list(from_waypoint)
-            self._current_waypoint_snapshots = dict()  # maps id to waypoint snapshot
-            self._current_edge_snapshots = dict()  # maps id to edge snapshot
-            self._current_annotation_name_to_wp_id = dict()
+            self._init_current_graph_nav_state()
 
             # Async Tasks
             self._async_task_list = []
@@ -1949,6 +1946,17 @@ class SpotWrapper:
 
     ###################################################################
 
+    def _init_current_graph_nav_state(self):
+        # Store the most recent knowledge of the state of the robot based on rpc calls.
+        self._current_graph = None
+        self._current_edges = dict()  # maps to_waypoint to list(from_waypoint)
+        self._current_waypoint_snapshots = dict()  # maps id to waypoint snapshot
+        self._current_edge_snapshots = dict()  # maps id to edge snapshot
+        self._current_annotation_name_to_wp_id = dict()
+        self._current_anchored_world_objects = dict() # maps object id to a (wo, waypoint, fiducial)
+        self._current_anchors = dict() # maps anchor id to anchor
+
+
     ## copy from spot-sdk/python/examples/graph_nav_command_line/graph_nav_command_line.py
     def _get_localization_state(self, *args):
         """Get the current localization and state of the robot."""
@@ -2035,7 +2043,7 @@ class SpotWrapper:
     def _upload_graph_and_snapshots(self, upload_filepath):
         """Upload the graph and snapshots to the robot."""
         self._logger.info("Loading the graph from disk into local storage...")
-        with open(upload_filepath + "/graph", "rb") as graph_file:
+        with open(os.path.join(upload_filepath, "graph"), "rb") as graph_file:
             # Load the graph from disk.
             data = graph_file.read()
             self._current_graph = map_pb2.Graph()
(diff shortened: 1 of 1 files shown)


spot_wrapper/wrapper.py
```diff
@@ -2035,7 +2044,7 @@ def _list_graph_waypoint_and_edge_ids(self, *args):
     def _upload_graph_and_snapshots(self, upload_filepath):
         """Upload the graph and snapshots to the robot."""
         self._logger.info("Loading the graph from disk into local storage...")
-        with open(upload_filepath + "/graph", "rb") as graph_file:
+        with open(os.path.join(upload_filepath, "graph"), "rb") as graph_file:
```
ggoretkin-bdai, line 2047: thanks for improving this!

spot_wrapper/wrapper.py
```diff
@@ -2047,25 +2056,56 @@ def _upload_graph_and_snapshots(self, upload_filepath):
...
+            with open(waypoint_filepath, "rb") as snapshot_file:
                 waypoint_snapshot = map_pb2.WaypointSnapshot()
                 waypoint_snapshot.ParseFromString(snapshot_file.read())
                 self._current_waypoint_snapshots[
                     waypoint_snapshot.id
                 ] = waypoint_snapshot
+
+                for fiducial in waypoint_snapshot.objects:
+                    if not fiducial.HasField("apriltag_properties"):
+                        continue
```
ggoretkin-bdai, line 2075: I encourage this style, instead of `if fiducial.HasField("apriltag_properties"):` and then nesting the rest of the body within.

spot_wrapper/wrapper.py
```diff
@@ -2047,25 +2056,56 @@ def _upload_graph_and_snapshots(self, upload_filepath):
...
+
+                    str_id = str(fiducial.apriltag_properties.tag_id)
+                    if (
+                        str_id in self._current_anchored_world_objects
+                        and len(self._current_anchored_world_objects[str_id]) == 1
+                    ):
```
ggoretkin-bdai, line 2081: and I think you should use the same style here too. Negate the conditional, replace the body with `continue`, and dedent the remainder.

spot_wrapper/wrapper.py
```diff
@@ -1949,6 +1946,18 @@ def grasp_3d(self, frame, object_rt_frame):
...
+        # Store the most recent knowledge of the state of the robot based on rpc calls.
+        self._current_graph = None
+        self._current_edges = dict()  # maps to_waypoint to list(from_waypoint)
+        self._current_waypoint_snapshots = dict()  # maps id to waypoint snapshot
+        self._current_edge_snapshots = dict()  # maps id to edge snapshot
+        self._current_annotation_name_to_wp_id = dict()
+        self._current_anchored_world_objects = (
+            dict()
+        )  # maps object id to a (wo, waypoint, fiducial)
+        self._current_anchors = dict()  # maps anchor id to anchor
```
ggoretkin-bdai, line 1959: I like the convention `_current_anchors_by_id`, which IMO removes the need for a comment.`. I suppose more explicit would be `_current_anchors_by_anchor_id`

spot_wrapper/wrapper.py
```diff
@@ -2047,25 +2056,56 @@ def _upload_graph_and_snapshots(self, upload_filepath):
...
+                upload_filepath, "edge_snapshots", edge.snapshot_id
+            )
+            if not os.path.exists(edge_filepath):
+                continue
+            with open(edge_filepath, "rb") as snapshot_file:
                 edge_snapshot = map_pb2.EdgeSnapshot()
                 edge_snapshot.ParseFromString(snapshot_file.read())
                 self._current_edge_snapshots[edge_snapshot.id] = edge_snapshot
+        for anchor in self._current_graph.anchoring.anchors:
+            self._current_anchors[anchor.id] = anchor
```
ggoretkin-bdai, line 2104: I want to be sure that it is correct to mutate this (possibly leaving "state" anchor ids from the last time it ran) instead of creating `_current_anchors` from scratch with juts the new information.

spot_wrapper/wrapper.py
kaiyu-zheng: This is actually pasted from BD's example code.

spot_wrapper/wrapper.py
kaiyu-zheng: I think in this case because the block is at the end of the for loop, it is still readable? Also this for loop is pasted from BD's example code. Perhaps leaving it this way makes that obvious

spot_wrapper/wrapper.py
kaiyu-zheng: Good point. I think in the context of spot_ros2, just looking at `id` one wouldn't know which id it is...

spot_wrapper/wrapper.py
kaiyu-zheng: The code again comes from BD example. The functionality that uses current_anchors hasn't been added yet. It is useful for informing what the frame of reference is for the point cloud map.

ggoretkin-bdai:
@kaiyu-zheng I think I am missing too much context to do an effective review. Here is a summary of my understanding of this PR

- has copied code from a BD example.
- introduces functionality that is not yet used.
- from the title, it checks edge caess, (but I don't understand what those edge cases are).

Let me know how I can be helpful.

kaiyu-zheng:
I am in the process of adding more functionality on the GraphNav side. This PR contains a single service that's a first step to enable other functionality (you need to upload the graph first in order to do things with it). Just wanted to put it out there for review, and possibly merge this as it has been tested. The next PR may involve several new services.

amessing-bdai (approved):

heuristicus:
Please apply black formatting before we merge.

amessing-bdai (approved):

PR #16, opened 2023-05-31
Updated bosdyn to 3.2.3

davidwatkins-bdai:
Updating the version of `bosdyn` to 3.2.3 as upstream dependencies are breaking `bosdyn-choreography-client`.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,3 +1,3 @@
-bosdyn_core==3.2.2.post1
+bosdyn_core==3.2.3
 protobuf==4.22.1
 setuptools==45.2.0


amessing-bdai (approved):
I don't see any issue, but this is definitely one where we need @heuristicus to weight in too

heuristicus:
No objections from me. It would be useful to discuss how we want to handle updates though. In practice I've not noticed problems upgrading to new versions of the SDK over the last couple of releases.

davidwatkins-bdai:
I tested this update already, so there should be no conflicts. We definitely need better update practices but for the time being manual updates are easiest.

davidwatkins-bdai:
Can one of you approve the PR so I can merge?

PR #17, opened 2023-06-01
[OC-4] Build a Spot Dance Interface

Guptabot:
Added execute_dance functionality to Spot Wrapper


heuristicus:
In the interest of the coming modularity changes, would you consider making this a separate file and class (see #6)? I'd rather not add more stuff which will need to be immediately modified by that PR.

spot_wrapper/spot_dancing.py
```diff
@@ -0,0 +1,147 @@
...
+        self._robot_name = robot_name
+        self._username = username
+        self._password = password
+        self._logger = logger
+
+        sdk = bosdyn.client.create_standard_sdk("UploadChoreography")
+        sdk.register_service_client(ChoreographyClient)
+        self._robot = sdk.create_robot(self._hostname)
+        if not self._robot:
+            print("create robot failed!")
```
heuristicus, line 36: These prints should use the logger.

spot_wrapper/spot_dancing.py
```diff
@@ -0,0 +1,147 @@
...
+            sys.exit(1)
+        lease_client = self._robot.ensure_client(LeaseClient.default_service_name)
+        lease = lease_client.acquire()
+        lk = LeaseKeepAlive(lease_client)
+        self._choreography_client = self._robot.ensure_client(
+            ChoreographyClient.default_service_name
+        )
+
+    @staticmethod
+    def authenticate(robot, username, password, logger):
```
heuristicus, line 57: Import the authenticate method from the wrapper rather than copying the code.

amessing-bdai: Is there a reason this should authenticate separately and not just use the `robot` variable created in SpotWrapper similar to the modules in #6 ?

heuristicus: Probably not. At the moment the changes indicate a standalone script, what I meant when suggesting modularity was to have it included in the wrapper initialisation like other modules in the PR you mentioned. It should probably print a short info message to indicate whether choreography is available or not.

amessing: Agree about it being a standalone script at the moment. The question about using the `robot` variable was more directed at the author of the PR. @Guptabot

Guptabot: There's no need for the variables to be separate. I'm working on changing the code accordingly now!

Guptabot: Should I also initialise self._robot_clients in wrapper.py (the way it is done in #6) then?

spot_wrapper/spot_dance.py
```diff
@@ -0,0 +1,83 @@
...
+        self._robot = robot
+        self._choreography_client = choreography_client
+        self._lease_client = lease_client
+        self._is_licensed_for_choreography = is_licensed_for_choreography
+
+    def execute_dance(self, filepath: str) -> tuple[bool, list]:
+        if not self._is_licensed_for_choreography:
+            return False, "Robot is not licensed for choreography."
+        lease = self._lease_client.acquire()
+        lk = LeaseKeepAlive(self._lease_client)
```
amessing-bdai, line 30: Is there a reason you need the lease_client or to create a LeaseKeepAlive? This seems like it would conflict with SpotWrapper getting a lease

spot_wrapper/spot_dance.py
```diff
@@ -0,0 +1,83 @@
...
+from bosdyn.client.lease import LeaseKeepAlive
+from bosdyn.client.robot import Robot
+import typing
+
+
+class SpotDance:
+    def __init__(
+        self,
+        robot: Robot,
+        choreography_client: typing.Any,
```
amessing-bdai, line 17: Can you put the actual [type](https://github.com/boston-dynamics/spot-sdk/blob/master/python/bosdyn-choreography-client/src/bosdyn/choreography/client/choreography.py) instead of Any

amessing-bdai:
No, need to create the robot_params/client stuff. Leave that to #6 as that is not what this PR is about.

spot_wrapper/spot_dance.py
```diff
@@ -0,0 +1,83 @@
...
+        lease_client: typing.Any,
+        is_licensed_for_choreography: typing.Any,
```
amessing-bdai, line 19: Also, can you put the actual type here?

spot_wrapper/spot_dance.py
Guptabot: I just tested this out after removing lines 29 and 30 and the code complains about not having a lease over the body. I was originally using @try_claim defined in wrapper.py. Would there be a conflcit with SpotWrapper given that I'm using the  same lease_client?

amessing-bdai: I would recommend having a dance function in SpotWrapper, put the decorator on that, and internal to that running your function from spot_dance. This way we avoid the circular import and also avoid trying to get a lease in multiple places.

spot_wrapper/spot_dance.py
```diff
@@ -0,0 +1,82 @@
...
+from bosdyn.choreography.client.choreography import ChoreographyClient
+from bosdyn.client.lease import LeaseClient
+
+
+class SpotDance:
+    def __init__(
+        self,
+        robot: Robot,
+        choreography_client: ChoreographyClient,
+        lease_client: LeaseClient,
```
amessing-bdai, line 18: Don't need the lease_client parameter anymore and then I think you are good to go (following a hardware test)

spot_wrapper/spot_dance.py
```diff
@@ -0,0 +1,82 @@
...
+        is_licensed_for_choreography: bool,
+    ):
+        self._robot = robot
+        self._choreography_client = choreography_client
+        self._lease_client = lease_client
+        self._is_licensed_for_choreography = is_licensed_for_choreography
+
+    def execute_dance(self, filepath: str) -> tuple[bool, list]:
```
davidwatkins-bdai, line 26: return type is `tuple[bool, str]`

davidwatkins-bdai (commented):
One typing issue, and address @amessing-bdai's issues.

davidwatkins-bdai (approved):

PR #18, opened 2023-06-03
fix crash on startup if choreography is not installed

heuristicus:
I don't know how it is that choreography is supposed to be installed (via pip?) but if it's not installed then the startup will crash. I assume most users won't install it. This PR handles that case.

Also makes sure the manipulation client is initialised to none if there is no arm. This will still lead to crashes if you try to call manipulation commands.

I've also included this in #6.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -783,6 +783,7 @@ class SpotWrapper:
                             self._choreography_client = None
                     else:
                         self._choreography_client = None
+                        self._is_licensed_for_choreography = False
 
                     try:
                         self._point_cloud_client = self._robot.ensure_client(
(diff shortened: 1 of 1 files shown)


amessing-bdai:
Is this ready for review?

heuristicus:
Yes, I think so.

amessing-bdai (approved):

Guptabot (approved):

heuristicus:
Non-BDAI users (like me) don't have access to that docker, and currently the instructions for the drivers don't include the choreography module. I suspect most people who are interested in using ROS won't want to use or have access to the choreography stuff, so leaving it out of installation is probably fine.

Mr-vedant-gupta:
Ah, I see

PR #23, opened 2023-06-09
FeatureBranch scientific notation for large numbers

philipsgithub:
as soon as numbers are 1e3 or bigger, counting the zeros becomes a thing. you can avoid that by using scientific notations

the scientific notation comes as a float by default so the corresponding type conversions in the code are adjusted as well

### DIFF
--- a/spot_wrapper/cam_wrapper.py
+++ b/spot_wrapper/cam_wrapper.py
@@ -227,7 +227,7 @@ class HealthWrapper:
             Tuple of string and float indicating the component and its temperature in celsius
         """
         return [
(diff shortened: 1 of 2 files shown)


philipsgithub:
it also matches style of the given maximum command duration in https://github.com/bdaiinstitute/spot_wrapper/blob/20fb44748e19c66284d55619fc893a418f5662b1/spot_wrapper/wrapper.py#L70

philipsgithub:
i formatted it with an updated black and still everything seems fine

amessing-bdai (approved):

PR #29, opened 2023-06-24
Set default args to list_graph()

sktometometo:
Set default args to list_graph() so that it can be called without setting args since that arg is not used.


kaiyu-zheng (approved):
LGTM. Not sure why it was like that.

sktometometo:
Thanks!

PR #32, opened 2023-06-28
Move initialisation of wrapper out of if statement

heuristicus:
This inverts the if statement checking whether the robot is initialised to checking if it is not. If it isn't, then the wrapper is invalid and returns without doing any of the initialisation. This reduces the indentation of the code and removes a massive if block.

For some reason the diff of this looks very weird, but the only thing I did here is to dedent lines 735 to 975.

This change comes from the modularity PR, but does not include any changes beyond the indentation.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -728,250 +728,251 @@ class SpotWrapper:
             self._valid = False
             return
 
-        if self._robot:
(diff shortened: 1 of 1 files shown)


amessing-bdai (approved):

amessing-bdai:
Tested

PR #33, opened 2023-06-28
Add typing to some wrapper functions

heuristicus:
Further non-modular changes to the wrapper which come from the modular PR. This adds typing to some functions in the wrapper which will not be moved by the modularity changes.

Also moves the SPOT_CLIENT_NAME from inside the wrapper to a global variable like MAX_COMMAND_DURATION, and simplifies the initialisation of the rates and callbacks dicts.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -13,8 +13,11 @@ import bosdyn.client.auth
 from bosdyn.api import arm_command_pb2
 from bosdyn.api import geometry_pb2
 from bosdyn.api import image_pb2
+from bosdyn.api import lease_pb2
+from bosdyn.api import point_cloud_pb2
 from bosdyn.api import manipulation_api_pb2
 from bosdyn.api import robot_command_pb2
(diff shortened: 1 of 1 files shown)


spot_wrapper/wrapper.py
```diff
@@ -1075,41 +1072,41 @@ def hand_images(self):
...
-    def near_goal(self):
+    def near_goal(self) -> bool:
         return self._near_goal
 
     @property
-    def at_goal(self):
+    def at_goal(self) -> bool:
         return self._at_goal
 
-    def is_estopped(self, timeout=None):
```
amessing-bdai, line 1105: timeout is an `typing.Optional[float]`. can you please add?

amessing-bdai (changes requested):
Will approve as soon as the requested changes are in. Sorry to be so pedantic, but it is a typehinting PR

heuristicus:
OK, I think I've covered pretty much everything now, including functions which don't return anything, and added some comments as wel.

amessing-bdai (approved):

PR #34, opened 2023-06-28
Module for world object retrieval

heuristicus:
This is the smallest module from the modular PR, with very few changes required.


heuristicus:
Is there a benefit to having type hints on `__init__`? As far as I know it's always expected to return None, and shouldn't ever be directly called by anyone.

heuristicus:
Seems like https://peps.python.org/pep-0484/ has the answer for `__init__` annotation.

amessing-bdai:
I think eventually we will try to add mypy like we have in spot_ros2 and it wants the return typehint on `__init__`, so this future proofs it for that. And yes, the reason mypy wants it is for what you found in pep-484

amessing-bdai (approved):

amessing-bdai (approved):

PR #35, opened 2023-06-28
Add CODEOWNERS file for visibility

amessing-bdai:
This is more for visibility than marking ownership. If someone external to one of our organizations makes a PR, we will automatically be put on as reviewers and can point the correct person to review it (if it is not one of us). Who else should be added to the list?


heuristicus (approved):
I don't think there's anyone else to include as a code owner at the moment.

jbarry-bdai (approved):

PR #37, opened 2023-06-28
Module for EAP pointclouds

heuristicus:
Creates the module for spot EAP pointcloud retrieval. In theory this could retrieve pointclouds from other internal sources as well.


spot_wrapper/spot_eap.py
```diff
@@ -0,0 +1,91 @@
...
+        super(AsyncPointCloudService, self).__init__(
+            "robot_point_cloud_service", client, logger, period_sec=1.0 / max(rate, 1.0)
+        )
+        self._callback = None
+        if rate > 0.0:
+            self._callback = callback
+        self._point_cloud_requests = point_cloud_requests
+
+    def _start_query(self) -> typing.Optional[FutureWrapper]:
+        if self._callback and self._point_cloud_requests:
```
amessing-bdai, line 43: Can these be explicit checks for `is not None` and an empty list?

spot_wrapper/spot_eap.py
```diff
@@ -0,0 +1,91 @@
...
+
+    """
+
+    def __init__(
+        self,
+        client: PointCloudClient,
+        logger: logging.Logger,
+        rate: float,
+        callback: typing.Callable,
+        point_cloud_requests: PointCloudRequest,
```
amessing-bdai, line 23: This should be `List[PointCloudRequest]`

spot_wrapper/wrapper.py
```diff
@@ -1014,6 +985,11 @@ def robot_name(self) -> str:
     def frame_prefix(self) -> str:
         return self._frame_prefix
 
+    @property
+    def spot_eap_lidar(self) -> SpotEAP:
```
amessing-bdai, line 989: Should be `Optional[SpotEAP]`

spot_wrapper/spot_eap.py
amessing-bdai, line 23: This should be `List[PointCloudRequest]`

spot_wrapper/spot_eap.py
amessing-bdai, line 43: Can these be explicit checks for `is not None` and an empty list?

spot_wrapper/wrapper.py
amessing-bdai, line 989: Should be `Optional[SpotEAP]`

amessing-bdai (changes requested):
Tested. Will approve after these changes are made.

spot_wrapper/spot_eap.py
heuristicus: What's your preferred method for a non-empty list check?  Just checking that `len` is nonzero? I tend to check lists the way it's done here since otherwise it's not very nice, and we're assuming we have a list anyway.

amessing-bdai: Yes, check if len is non-zero. I agree that it isn't the prettiest, but recently had to fix a number of bugs caused by `if <var>` handling `None` and empty strings/lists, so currently feel that the explicit check is safer.

heuristicus: I went a little overboard and added a type check as well as a length check. Probably don't need the length check if we're checking that it's the correct type, but doesn't hurt either.

amessing-bdai (approved):

PR #39, opened 2023-06-28
Module for autonomous docking

heuristicus:
Moves autonomous docking into its own module.

- [x] This requires working out how to store the robot parameters like is_sitting. The previous approach in the modular PR doesn't work because the dictionary is static once initialised.
- [x] The docking state has to set variables on the main wrapper for the docking command
- [ ] Regression due to `try_claim` no longer being available

Requires #58 to be merged.


philipsgithub:
Hi @heuristicus, could you explain the advantages of this PR? i saw that you want to modularize several components and wonder what benefits you are aiming for. I can think of a few but you probably have a deeper understanding:)

heuristicus:
As you can see with the many other PRs I've made recently, this comes from an effort to split the responsibility for handling certain components of the system out to specific classes. The current wrapper is monolithic and so it's quite hard for anyone looking at the code to understand what things affect what other things, and the methods are all in one place. The aim is to make the wrapper easier to maintain and modify.

jbarry-bdai (commented):
This will need a corresponding change in spot_ros2.  @mpickett-bdai can you handle that change, update, and testing that the dock and undock functions still work?

mpickett-bdai:
I was attempting to test this PR on some of our Spots here and ran into some build issues. The spot_docker.py imports the `wrapper` in a way that isn't fully visible on a package layer. Using cam_wrapper.py as a model we call the import with `spot_driver.wrapper`. Additionally when I tried this fix on my end for testing, we get a circular dependency issue since wrapper.py calls spot_docker.py and vice-versa. Maybe we have to separate it like we have with cam_wrapper?

heuristicus:
I don't know what the best way to avoid this sort of thing is, but probably the easiest way is to move the classes we are importing from the wrapper into a "helpers" file and import that into both the wrapper and the modules. None of the modules should need to import the wrapper itself.

heuristicus:
Draft PR at https://github.com/bdaiinstitute/spot_wrapper/pull/58, you should be able to do

```
from spot_wrapper.wrapper_helpers import RobotCommand, RobotState
```

Or something along those lines.

mpickett-bdai:
I'll give this a shot some time tomorrow, but I think thats a good path forward.

spot_wrapper/wrapper.py
```diff
@@ -908,6 +909,14 @@ def __init__(
             self._estop_monitor,
         ]
 
+        self._spot_docking = SpotDocking(
+            self._robot,
+            self._logger,
+            self._state,
+            self._docking_client,
```
mpickett-bdai, line 916: Missing RobotCommandData

mpickett-bdai: Otherwise with the cherry pick of PR #58, the above fix and the added import of `from spot_wrapper.wrapper_helpers import RobotCommandData, RobotState`  into the spot_docking.py seems to work with the appropriate changes to spot_ros2. @heuristicus Go head and make the above fixes and put #58 up for review. Once its all together I can do another quick test to make sure it all works.

mpickett-bdai (changes requested):

mpickett-bdai:
@heuristicus Tested out the latest and it works well :). Just needs to be updated to the main branch, which I think might have some conflicts since the last time I tried it locally.

mpickett-bdai (approved):

mpickett-bdai (approved):
Tested after updated to latest. LGTM

PR #41, opened 2023-06-29
Store robot state in a dataclass to pass to modules

heuristicus:
This is part of what is required for the module changes but affects how I implement those.

Some of the modules need access to the state of the robot's sitting/standing/moving state to determine their behaviour. In the original modular PR this was implemented as a dict that would be passed to the modules, but I think an object is a better approach.

This dataclass can be passed to modules and when its attributes are changed by async_idle in the wrapper, the changes should be visible in the modules.

I'm not certain this is the best way to do this, but it's probably good enough for now.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -351,21 +351,28 @@ class AsyncPointCloudService(AsyncPeriodicQuery):
 
 
 class AsyncIdle(AsyncPeriodicQuery):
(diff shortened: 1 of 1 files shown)


amessing-bdai:
Looks good. Will run a test later today when I get a chance

amessing-bdai (approved):
Tested

amessing-bdai (approved):
Tested

PR #42, opened 2023-06-29
Module for image retrieval

heuristicus:
This breaks out image retrieval into its own module.

- [x] Currently this includes the hand image, and the hand image is also separately retrieved by the arm module. Need to work out where that code should actually be. The two work slightly differently, the arm uses the old async image stuff to get images, I think. Probably just use the image module?


heuristicus:
- [ ] Currently this includes the hand image, and the hand image is also separately retrieved by the arm module. Need to work out where that code should actually be. The two work slightly differently, the arm uses the old async image stuff to get images, I think.

amessing-bdai:
Yeah, I would think hand image stuff goes in the image module.

spot_wrapper/wrapper.py
```diff
@@ -5,14 +5,10 @@
 import time
 import traceback
 import typing
-from enum import Enum
-from collections import namedtuple
-from dataclasses import dataclass, field
```
bhung-bdai, line 10: Apologies for the late response. You still need to import the dataclass because https://github.com/heuristicus/spot_wrapper/blob/be0e81e551f29f9ba7a43642e304831374478f9a/spot_wrapper/wrapper.py#L426 references it. Other than that, LGTM.

bhung-bdai (changes requested):

spot_wrapper/wrapper.py
```diff
@@ -1,18 +1,15 @@
+from dataclasses import dataclass
```
bhung-bdai, line 1: Added the dataclass import. Works as expected in testing with Rviz otherwise.

bhung-bdai (approved):

amessing-bdai (changes requested):
Merge conflict with main needs to be resolved

heuristicus:
Resolved the conflict, and made some small changes to reduce code duplication and make it possible to specify the image quality you want. This isn't in the wrapper code yet but we can integrate that later.

heuristicus:
@bhung-bdai thanks for catching the missing import, it must have got lost somewhere in the process of making the new PR for the EAP.

amessing-bdai (approved):

PR #43, opened 2023-06-29
Module for arm

heuristicus:
- [x] Requires robot params access
- [x] Work out what to do with the hand image
- [x] Work out how to call blocking stand, and what to do with the robot_command parameter
- [ ] try_claim regression


spot_wrapper/spot_arm.py
```diff
@@ -0,0 +1,609 @@
...
+        try:
+            success, msg = self.ensure_arm_power_and_stand()
+            if not success:
+                self._logger.info(msg)
+                return False, msg
+            else:
+                # The open angle command does not take degrees but the limits
+                # defined in the urdf, that is why we have to interpolate
+                closed = 0.349066
+                opened = -1.396263
```
amessing-bdai, line 431: This is interesting because the [robot command builder](https://github.com/boston-dynamics/spot-sdk/blob/master/python/bosdyn-client/src/bosdyn/client/robot_command.py#L32) uses:
`_CLAW_GRIPPER_OPEN_ANGLE = -1.5708
_CLAW_GRIPPER_CLOSED_ANGLE = 0
`

heuristicus: These changes look to have been made by @maubrunn, not sure why they are different.

spot_wrapper/spot_arm.py
```diff
@@ -0,0 +1,609 @@
...
+    RobotCommandBuilder,
+    RobotCommandClient,
+    block_until_arm_arrives,
+    blocking_stand,
+)
+from bosdyn.client.robot_state import RobotStateClient
+from bosdyn.client.time_sync import TimeSyncEndpoint
+from bosdyn.util import seconds_to_duration
+
+from wrapper import RobotState
```
bhung-bdai, line 23: This is showing up as a circular dependency when I try to run it. wrapper.py imports spot_arm.py and spot_arm.py imports wrapper.py.

bhung-bdai (changes requested):

spot_wrapper/spot_arm.py
```diff
@@ -0,0 +1,609 @@
...
+                # Send the request
+                cmd_id = self._robot_command_client.robot_command(robot_command)
+                self._logger.info("Force trajectory command sent")
+                self.wait_for_arm_command_to_complete(cmd_id)
+        except Exception as e:
+            return False, f"Exception occured during arm movement: {e}"
+
+        return True, "Moved arm successfully"
+
+    def gripper_open(self) -> typing.Tuple[bool, str]:
```
bhung-bdai, line 358: When I run `gripper_open, gripper_close, gripper_angle_open` I get an error while the system gets stuck on `block_until_arm_arrives` forever. Is this just an error on my side? the `arm._unstow, arm_carry, arm_stow` functions don't have the same issue.

amessing-bdai: `block_until_arm_arrives` doesn't have any checks for gripper commands

bhung-bdai (changes requested):

spot_wrapper/spot_arm.py
heuristicus: Looks like I need to use https://dev.bostondynamics.com/protos/bosdyn/api/proto_reference.html#clawgrippercommand-feedback-status and write something similar to the blocking manipulation command.

heuristicus:
Gripper commands should now work properly. I've made it so that the command blocks either until the goal is reached, or until a force is being applied by the gripper. This means that the `close` and `angle_open` commands should both terminate early if the claw is gripping something.

bhung-bdai (approved):

amessing-bdai (approved):

bhung-bdai (approved):

PR #48, opened 2023-06-30
Dataclass for storing command IDs

heuristicus:
The `last_*_command` attributes of the wrapper are associated with each other, but not in any proper way. This adds a dataclass to hold these attributes instead.

Also adds properties to get/set the values of these attributes. This means that `AsyncIdle` is no longer accessing protected members of the wrapper.

This is also required for modularity purposes. The docking module sets the docking command information, and other modules may want to set other attributes.

`_last_robot_command_feedback` was unused so I removed it.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -373,10 +373,10 @@ class AsyncIdle(AsyncPeriodicQuery):
         self._spot_wrapper: SpotWrapper = spot_wrapper
 
     def _start_query(self) -> None:
-        if self._spot_wrapper._last_stand_command != None:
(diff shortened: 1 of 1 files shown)


sktometometo (changes requested):
https://github.com/heuristicus/spot_wrapper/blob/557b1c319915eca15265818e53d32b72b71f53de/spot_wrapper/wrapper.py#L446
This line should also be updated.

amessing-bdai (approved):
Looks good. @heuristicus Is this ready to merge?

heuristicus:
Yes, this should be good to merge.

PR #49, opened 2023-07-02
Implement MediaLog functions for spot cam

heuristicus:
The ImageStreamWrapper does not allow you to save high quality images - it's supposed to be used for reference only as the images from webrtc have a lot of compression applied. This PR provides access to the MediaLog commands, which save images to disk on the cam and can then be read and saved locally once they are processed. This should give access to the highest quality images.

- [x] Test with an actual spot cam

### DIFF
--- a/spot_wrapper/cam_wrapper.py
+++ b/spot_wrapper/cam_wrapper.py
@@ -450,9 +450,9 @@ class MediaLogWrapper:
         """
         # TODO: What does this return if the logpoint doesn't exist? Just an empty logpoint?
         if raw:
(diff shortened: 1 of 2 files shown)


heuristicus:
I've tested this on a robot with a spot cam. The saving functionality is a bit weird but it's adapted from the BD examples and works. Can be improved later if necessary.

heuristicus:
This is ready for review/testing/merge. We've had a collaborator's robot running this code for a while and saving images with it.

amessing-bdai (approved):

PR #50, opened 2023-07-05
[N/A] Remove typevar from spot_world_objects.py

amessing-bdai:
We added the typevar because using a protobuf enum type in a typing.Generic (e.g. typing.List) throws an exception during runtime: `TypeError: Parameters to generic types must be types. Got <google.protobuf.internal.enum_type_wrapper.EnumTypeWrapper object`.

You can remove the TypeVar if you use the annotations module from __future__ as explained [here](https://stackoverflow.com/questions/60342896/how-to-use-type-hinting-with-dictionaries-and-google-protobuf-enum). This simply postpones the evaluation of annotations.

### DIFF
--- a/spot_wrapper/spot_world_objects.py
+++ b/spot_wrapper/spot_world_objects.py
@@ -1,15 +1,11 @@
+from __future__ import annotations
 import logging
 import typing
 
-from bosdyn.api.world_object_pb2 import ListWorldObjectResponse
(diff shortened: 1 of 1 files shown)


jbarry-bdai:
Did you test this on robot?

amessing-bdai:
Yes

Issue #51, opened 2023-07-05
Implement `try_claim` for modules


heuristicus:
As mentioned in #44 and #39, some of the modules have slight regression because the `try_claim` decorator is linked to access to some parts of the main wrapper object. It would be good to have an object or some other mechanism that can be used to allow modules to make use of the functionality in a similarly transparent way.

Updates needed in:
- [ ] Graphnav module
- [ ] Docking module. Done in #55.
- [ ] Arm module

PR #55, opened 2023-07-08
Try-claim decorator converted to an object

heuristicus:
Currently the `try_claim` decorator is a function which is expected to function using the `@` decorator syntax. It is also intrinsically linked to the `SpotWrapper` object, which means the new modules cannot make use of it. Many of the functions making use of the decorator should be in modules, so there needs to be a way of getting the decorator to the modules.

This PR implements the decorator as a method on an object. The `TryClaimDecorator` object is instantiated by the wrapper. The methods which previously used the `@` syntax are now decorated during the instantiation of the wrapper. The same mechanism can be applied to methods of other objects.

I also took the opportunity to clean up the decorator a little bit, it should be much clearer what it's actually doing now.

Should be the fix for #51.

### DIFF
--- a/spot_wrapper/spot_arm.py
+++ b/spot_wrapper/spot_arm.py
@@ -21,7 +21,7 @@ from bosdyn.client.robot_state import RobotStateClient
 from bosdyn.client.time_sync import TimeSyncEndpoint
 from bosdyn.util import seconds_to_duration
 
-from spot_wrapper.wrapper_helpers import RobotState
+from spot_wrapper.wrapper_helpers import RobotState, ClaimAndPowerDecorator
 
 
 class SpotArm:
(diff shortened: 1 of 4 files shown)


jbarry-bdai:
I'm going to let @amessing-bdai review this one when he gets back as he wrote the original decorator...

amessing-bdai:
Can you refactor this for main with the helper file? Also, I think this class should go in the helper file in case we want to pass an instance of it to a module and need the typehint

heuristicus:
Updated to include changes to main, and moved to the helper file. I updated comments and some names to try and make it more obvious what the decorator is for.

I couldn't work out what type hint to use for "some instantiated object". I could use `typing.Any` but that includes base types like `int`, which the function would not work for. I could be very specific and write a union of the wrapper, docking, and whatever else we will eventually apply it to, but that doesn't feel right either.

amessing-bdai:
Maybe a typevar with a comment would be best. A union of the wrapper and others leaves the potential for the typehint to not be updated if a new module is added in the future.

Also, the PR is currently adding RobotState and RobotCommandData back to wrapper.py

bhung-bdai (approved):

PR #58, opened 2023-07-11
Helper file for classes defined in wrapper and used in modules

heuristicus:
For correct type hints, some modules have to import the dataclasses used by the wrapper. This causes circular imports. This PR moves the dataclasses out into a helper file to prevent this.


mpickett-bdai:
Yeah I like this. It makes everything cleaning and allows us to expand if there are more shared data classes in the future.

amessing-bdai (approved):

PR #59, opened 2023-07-22
Break wrapper parameters out into a dataclass

heuristicus:
The wrapper currently has 13 parameters to its init function, which is too many. The wrapper may need to receive other configuration information over time, and rather than adding to the parameters of the function, it would be easier and break fewer things if we did it another way.

In this PR I've taken the parameters and put them into a dataclass which the wrapper takes instead.


spot_wrapper/wrapper.py
```diff
@@ -411,63 +411,61 @@ def wrapper_try_claim(self, *args, **kwargs):
     return wrapper_try_claim
 
 
+@dataclass()
```
ksharma-bdai, line 414: should this just be `@dataclass` rather than `@dataclass()` - i haven't seen the latter syntax before

spot_wrapper/wrapper.py
```diff
@@ -411,63 +411,61 @@ def wrapper_try_claim(self, *args, **kwargs):
...
+    rates: typing.Optional[typing.Dict[str, float]] = None
+    callbacks: typing.Optional[typing.Dict[str, typing.Callable]] = None
+    use_take_lease: bool = False
+    get_lease_on_action: bool = False
+    continually_try_stand: bool = True
+    rgb_cameras: bool = True
+    payload_credentials_file: str = None
+    frame_prefix: str = field(init=False)
+
+    def __post_init__(self) -> None:
```
ksharma-bdai, line 454: For consistency with the old behavior, `__post_init__` does this also need to handle these cases from the original?

```
self._rates = rates or {}
self._callbacks = callbacks or {}
```

Otherwise I believe all of the `.get` calls below on `rates` and `callbacks` will fail for `NoneType` in all of the different tasks.

Or maybe just initialize them to empty lists instead of `None`?

ksharma-bdai (changes requested):
Purely a semantic code review, have not tested on robot, but I think a few small changes are needed for consistency between the config class and old initialization. Let me know if you disagree.

spot_wrapper/wrapper.py
heuristicus: I'm not sure if dataclasses work in the same way as mutable initialisation arguments, if they do that would cause some issues. I've initialised in post-init.

ksharma-bdai:
LGTM - This will require a corresponding change downstream in `spot_ros2` - I'll put up that PR within a day or two during hw testing and approve this.

ksharma-bdai:
@heuristicus I've been thinking about this some more while looking at the `spot_ros2` change required and I'm a bit undecided if this change is for the best. I do like using parameter classes in general, but in this case:

Pros
- 13 parameters is a lot, this separates the class out cleanly with more documentation

(Potential) Cons
- we really only initialize `SpotWrapper` once in one place
- unless we change the downstream code to internally reference this config, we're building a config once for the wrapper and then continuing to reference the class's own internal members instead
- we may actually want callers to break if a new parameter is added rather than silently ignoring/being unaware of its existence

In either case it seems like new parameters have been added with default values, the only breaking updates for callers would be for non-optional positional parameters or changing the order of the arguments list.

The shortest path change in `spot_ros2` would simple be changing:
```
self.spot_wrapper = SpotWrapper(
    self.username,
    self.password,
    self.ip,
    self.name,
    self.wrapper_logger,
    self.start_estop.value,
    self.estop_timeout.value,
    self.rates,
    self.callbacks,
    self.use_take_lease.value,
    self.get_lease_on_action.value,
    self.continually_try_stand.value,
    self.rgb_cameras.value,
)
```

to:
```
self.spot_wrapper = SpotWrapper(
    SpotWrapperConfiguration(
        username=self.username,
        password=self.password,
        ip=self.ip,
        name=self.name,
        logger=self.wrapper_logger,
        start_estop=self.start_estop.value,
        estop_timeout=self.estop_timeout.value,
        rates=self.rates,
        callbacks=self.callbacks,
        use_take_lease=self.use_take_lease.value,
        get_lease_on_action=self.get_lease_on_action.value,
        continually_try_stand=self.continually_try_stand.value,
        payload_credentials_file=self.rgb_cameras.value
    )
)
```

which is not a lot different than changing the original code to simply use named rather than positional arguments to the 13 argument constructor.

To have proper downstream code reuse we'd need to change `spot_ros2` to internally use the configuration class and audit the codebase to make sure all of the appropriate references are updated (as you did in the wrapper properties to make the config essentially transparent to the "outside" world):

```
    @property
    def robot_name(self) -> str:
        return self._config.robot_name

    @property
    def frame_prefix(self) -> str:
        return self._config.frame_prefix
```

cc @amessing-bdai any thoughts on this as well?

baxelrod-bdai:
Notes from the peanut gallery: I think in general I am more in favor of using named or default parameters rather than a configuration class as @ksharma-bdai showed.

PR #62, opened 2023-07-28
[N/A] Fixed image quality typing

amessing-bdai:
Image quality in the protobuf is a double: https://github.com/boston-dynamics/spot-sdk/blob/master/protos/bosdyn/api/image.proto#L194C5-L194C11

- Needs HIL testing


heuristicus (approved):
Good catch.

PR #31, opened 2023-06-24
Fix logging output of graph_nav_util.py

sktometometo:
Fix logging args

PR #65, opened 2023-07-31
[SW-341] Fix incorrect edge logging format in graphnav_util

ksharma-bdai:
Incorrect logging format in `update_waypoints_and_edges` would result in a crash when attempting to invoke the `list_graph` endpoint.


sktometometo:
It seems related to https://github.com/bdaiinstitute/spot_wrapper/pull/31

ksharma-bdai:
> It seems related to #31

Oh thanks for pointing this out, I'll abandon this PR and we can wait for your fix to merge instead.

PR #66, opened 2023-08-05
Add claim decorator to graphnav module

heuristicus:
There are various bits where the lease is claimed independent of the try-claim, so that may have to be looked at in more detail, but this should mean graphnav functionality can be used without having to manually claim.


amessing-bdai:
@bhung-bdai @mpickett-bdai @kaiyu-zheng This should be testable with spot_ros2 by running the driver, hijacking the lease with the tablet, and seeing if spot_wrapper auto-recovers it

bhung-bdai (approved):
LGTM, works on my HIL test.

PR #70, opened 2023-08-22
Add aiortc to requirements txt

heuristicus:
Should fix https://github.com/bdaiinstitute/spot_ros2/issues/134.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,4 @@ bosdyn_core==3.2.3
 protobuf==4.22.1
 setuptools==45.2.0
 pytest==7.3.1
+aiortc
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,5 +1,8 @@
-bosdyn_core==3.2.3
+bosdyn-core==3.2.3
+bosdyn-client==3.2.3
+bosdyn-mission==3.2.3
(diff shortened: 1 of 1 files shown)


bhung-bdai:
I've only ever tested this with 1.5.0. Do you think a generic version works fine enough?

bhung-bdai (approved):
LGTM. Double checking, I think the versioning is fine but it might be worth making sure.

bhung-bdai (changes requested):
After some deliberation, I've decided to make some edits and edit `spot_ros2` to make the install a bit easier.

heuristicus:
We may need to be careful when specifying versions that the ones we pick are valid for both spot_ros2 and spot_ros, since the latter is running on the older LTS version. In this case I think it's OK.

heuristicus:
Although, should we not be using 3.3.0 for bosdyn now?

bhung-bdai (approved):
LGTM

bhung-bdai:
@heuristicus I should also add that the SDK versions go together. The file before my changes had a bosdyn-core version of 3.2.3. This should be requiring versions 3.2.3 of the other bosdyn packages.

amessing-bdai (approved):

PR #74, opened 2023-10-16
Introduce choreography 3.2.3 dependency

kkarol-bdai:
Add explicit dependency on choreography 3.2.3 
Remove get_status code that targets 3.3.0

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -43,22 +43,16 @@ from bosdyn.client.spot_check import SpotCheckClient
 from bosdyn.client.time_sync import TimeSyncEndpoint
 from bosdyn.client.world_object import WorldObjectClient
 
-try:
-    from bosdyn.choreography.client.choreography import (
-        ChoreographyClient,
-    )
-    from bosdyn.api.spot.choreography_sequence_pb2 import (
-        Animation,
-        ChoreographySequence,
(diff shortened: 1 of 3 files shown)


spot_wrapper/wrapper.py
```diff
@@ -436,15 +430,19 @@ def __init__(
...
 
         try:
             self._sdk = create_standard_sdk(SPOT_CLIENT_NAME)
         except Exception as e:
             self._logger.error("Error creating SDK object: %s", e)
             self._valid = False
             return
-        if HAVE_CHOREOGRAPHY:
+        try:
             self._sdk.register_service_client(ChoreographyClient)
```
kkarol-bdai, line 442: @ksharma-bdai we shouldn't merge this code until we test on a robot that doesn't have the choreography service.

kkarol-bdai: Functionality confirmed on robot without a license

spot_wrapper/wrapper.py
```diff
@@ -515,7 +513,7 @@ def __init__(
                 else:
                     self._gripper_cam_param_client = None
 
-                if HAVE_CHOREOGRAPHY:
+                try:
```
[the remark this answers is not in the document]
kkarol-bdai: Confirmed that this can  be safely removed on a robot wtithout a license

PR #77, opened 2023-10-31
Updated requirements to use ROS2 messages 3.3.2

gremigi-bdai:
In this PR I'm updating the requirements for the `spot_ros2` installation script to download and install the debian packages containing the ROS2 messages for Post SDK 3.3.2.

`spot_wrapper` is a submodule of `spot_ros2` package and this PR must be approved before the following:

https://github.com/bdaiinstitute/spot_ros2/pull/177

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,9 @@
-bosdyn-core==3.2.3
-bosdyn_choreography_client==3.2.3
-bosdyn_choreography_protos==3.2.3
-bosdyn-client==3.2.3
-bosdyn-mission==3.2.3
-bosdyn-api==3.2.3
+bosdyn-core==3.3.2
+bosdyn_choreography_client==3.3.2
+bosdyn_choreography_protos==3.3.2
+bosdyn-client==3.3.2
+bosdyn-mission==3.3.2
+bosdyn-api==3.3.2
 protobuf==4.22.1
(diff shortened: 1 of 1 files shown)


amessing-bdai (approved):

bhung-bdai (approved):
I will approve this but I think there's a significant amount of stakeholders to notify about this change.

jbarry-bdai:
Is this a breaking change?  I know the BD protos are supposed to be backwards compatible.

amessing-bdai:
They list the breaking changes in their [release notes](https://dev.bostondynamics.com/docs/release_notes). These are the only ones.

```
When the robot loses connection with an E-stop endpoint, an attempt to power on the robot will now return STATUS_KEEP_ALIVE_MOTOR_OFF instead of STATUS_ESTOPPED. In python code it will raise the KeepaliveMotorsOffError exception instead of EstoppedError.

Some malformed Graph Nav graphs were previously accepted, but would leave the service in a bad state. These graphs are now explicitly rejected during upload.
```

bhung-bdai:
> Is this a breaking change? I know the BD protos are supposed to be backwards compatible.

My very basic testing hasn't had issues so far, but it's by no means comprehensive.

amessing-bdai (approved):

PR #78, opened 2023-11-15
[N/A] Force merged main and changed requirement

bhung-bdai:
Updated the version of `setuptools` to a version that actually supports ROS2 as well.


amessing-bdai:
@heuristicus Does this cause any issues for you?

heuristicus:
I don't think this should be a problem, it's possible to get this version of setuptools on 20.04. I assume everything is backwards compatible.

amessing-bdai (approved):

PR #79, opened 2023-11-28
claim estop on power on only if isn't already acquired

skpawar1305:
For spot_driver with estop, if power_off is called, and power_on is called, driver crashes as an error is thrown as it tries to reset estop. Prevent this case.


bhung-bdai (approved):
HIL tested

PR #82, opened 2023-12-19
Spot CAM stuck in infinite loop when setting PTZ position

2b-t:
On the current `main` branch (commit 0c5e3c59d8658e6c15604a4626819031d5be7588) **`set_ptz_position`** with **`blocking`** set to **`True`** might result in the Spot CAM (tested in a Spot CAM+IR with [v3.3.0](https://bostondynamics.my.site.com/SupportCenter/s/article/Spot-CAM-3-x-Release-Notes)) never returning due to an incorrect implementation of the timeout:

https://github.com/bdaiinstitute/spot_wrapper/blob/0c5e3c59d8658e6c15604a4626819031d5be7588/spot_wrapper/cam_wrapper.py#L797-L802

If either `pan`, `tilt` or `zoom` do not converge, the code will be stuck in an infinite loop.

Furthermore the members of `current_position`, at least in my version of [`bosdyn.client.spot_cam.ptz.PtzClient`](https://dev.bostondynamics.com/_modules/bosdyn/client/spot_cam/ptz) (`3.3.2`), are all [**Google Protobuf `FloatValue`s**](https://googleapis.dev/python/protobuf/latest/google/protobuf/wrappers_pb2.html) and as a result I have to access their values with `.value` when doing comparisons. Not sure if this has always been the case as I am pretty sure that we had the same code running without any issues on `3.2.3`.

So the correct while loop condition that works for me is:

```python
while (
  (not math.isclose(current_position.pan.value, pan, abs_tol=1)
  or not math.isclose(current_position.tilt.value, tilt, abs_tol=1)
  or not math.isclose(current_position.zoom.value, zoom, abs_tol=0.5))
  and datetime.datetime.now() - start_time < datetime.timedelta(seconds=3)
)
```

### DIFF
--- a/spot_wrapper/cam_wrapper.py
+++ b/spot_wrapper/cam_wrapper.py
@@ -795,10 +795,10 @@ class PTZWrapper:
                 self._get_ptz_description(ptz_name)
             )
             while (
-                datetime.datetime.now() - start_time < datetime.timedelta(seconds=3)
-                or not math.isclose(current_position.pan, pan, abs_tol=1)
-                or not math.isclose(current_position.tilt, tilt, abs_tol=1)
-                or not math.isclose(current_position.zoom, zoom, abs_tol=0.5)
+                (not math.isclose(current_position.pan.value, pan, abs_tol=1)
+                or not math.isclose(current_position.tilt.value, tilt, abs_tol=1)
+                or not math.isclose(current_position.zoom.value, zoom, abs_tol=0.5))
+                and datetime.datetime.now() - start_time < datetime.timedelta(seconds=3)
             ):
                 current_position = self.client.get_ptz_position(
                     self._get_ptz_description(ptz_name)
--- a/spot_wrapper/cam_wrapper.py
+++ b/spot_wrapper/cam_wrapper.py
@@ -795,9 +795,9 @@ class PTZWrapper:
                 self._get_ptz_description(ptz_name)
             )
             while (
-                (not math.isclose(current_position.pan.value, pan, abs_tol=1)
-                or not math.isclose(current_position.tilt.value, tilt, abs_tol=1)
-                or not math.isclose(current_position.zoom.value, zoom, abs_tol=0.5))
+                not (math.isclose(current_position.pan.value, pan, abs_tol=1)
+                and math.isclose(current_position.tilt.value, tilt, abs_tol=1)
+                and math.isclose(current_position.zoom.value, zoom, abs_tol=0.5))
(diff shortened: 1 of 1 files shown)


spot_wrapper/cam_wrapper.py
```diff
@@ -795,10 +795,10 @@ def set_ptz_position(self, ptz_name, pan, tilt, zoom, blocking=False):
                 self._get_ptz_description(ptz_name)
             )
             while (
-                datetime.datetime.now() - start_time < datetime.timedelta(seconds=3)
-                or not math.isclose(current_position.pan, pan, abs_tol=1)
-                or not math.isclose(current_position.tilt, tilt, abs_tol=1)
-                or not math.isclose(current_position.zoom, zoom, abs_tol=0.5)
+                (not math.isclose(current_position.pan.value, pan, abs_tol=1)
```
bhung-bdai, line 798: As slight nit (but moreso to keep with the accepted code style), can you refactor the logic as
```
while (
    not (math.isclose(current_position.pan.value, pan, abs_tol=1) 
    and math.isclose(current_position.tilt.value, tilt, abs_tol=1)
    and math.isclose(current_position.zoom.value, zoom, abs_tol=0.5))
    and datetime.datetime.now() - start_time < datetime.timedelta(seconds=3)
```

I believe that logic is a bit easier to read and equivalent. This way, it's more obvious that we are exiting only in the case where each PTZ approaches the correct setting and the clock is under the convergence time.

bhung-bdai (commented):
Good catch. A slight nit but it looks good otherwise.

spot_wrapper/cam_wrapper.py
2b-t: Sure, I agree. Potentially we could also extract the timeout and put it as a conditional break inside the while loop body.

bhung-bdai (changes requested):
You may need to run the precommit as described on the home page of the repo to fix the formatting.

2b-t:
> You may need to run the precommit as described on the home page of the repo to fix the formatting.

As I was using the original ROS Noetic [`spot_ros`](https://github.com/heuristicus/spot_ros) driver, I was not aware that the wrapper was using the [pre-commit from `spot_ros2`](https://github.com/bdaiinstitute/spot_ros2/blob/3cac02b61e6b74192c85872b02f624fd5e065535/.pre-commit-config.yaml). Just ran it.
_Merry Christmas_

bhung-bdai (approved):

bhung-bdai:
No worries. Thanks for updating and have a good holiday season!

PR #83, opened 2024-01-08
Add property for accessing `SpotWrapper._lease`

ggoretkin-bdai:
We had a need to access the lease that results from the `SpotWrapper.getLease` method, but `lease` is already taken to access the proto of the lease that results from a periodic query via `AsyncLease`, hence `lease2`.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -30,7 +30,7 @@ from bosdyn.client.estop import (
 from bosdyn.client.graph_nav import GraphNavClient
 from bosdyn.client.gripper_camera_param import GripperCameraParamClient
 from bosdyn.client.image import ImageClient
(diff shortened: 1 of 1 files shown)


bhung-bdai (approved):
Weird. Do we actually need to access the lease? If possible, I would list the property as `spot_lease` or something instead of `lease2`

ashay-bdai:
The current `lease` property still has to do with spot, so I think another one called `spot_lease` would be confusing, suggesting that the `lease` property doesn't have to do with spot.

ggoretkin-bdai:
To echo @ashay-bdai , exposing the `_lease` field allows `SpotWrapper` users who need to share Spot between multiple clients to write some defensive code.

amessing-bdai (approved):

PR #81, opened 2023-12-06
Add `spot_wrapper.testing` machinery

mhidalgo-bdai:
This PR introduces `pytest` compatible machinery to aid in testing `spot_wrapper`, `spot_ros2`, and beyond.

PR #88, opened 2024-02-22
Allow repeated future outcomes in mocked gRPCs

mhidalgo-bdai:
Follow-up to #81. This patch should make it easier to work with blocking calls that invoke many gRPCs. 

Now, in addition to `future.returns()` and `future.fails()`, you can invoke `future.returns().repeatedly(N)` and `future.fails().repeatedly(N)` to repeat the same outcome a finite number of times and  `future.returns().forever()` and `future.fails().forever()` to repeat the same outcome indefinitely.

### DIFF
--- a/spot_wrapper/testing/grpc.py
+++ b/spot_wrapper/testing/grpc.py
@@ -353,7 +353,7 @@ class DeferredRpcHandler(GeneralizedDecorator):
 
             def resolve(self, call: "DeferredRpcHandler.Call") -> bool:
                 """Resolves the given `call` using this outcome."""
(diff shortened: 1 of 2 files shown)


mhidalgo-bdai:
Putting it out for you to try @jbarry-bdai. I'll add more tests and update the wiki later.

amessing-bdai (approved):

jbarry-bdai (approved):
Didn't look at the code but like the API!

mhidalgo-bdai:
OK, this is ready to go @amessing-bdai.

amessing-bdai (approved):

PR #92, opened 2024-03-07
Manage leasing resources properly

mhidalgo-bdai:
This patch changes ensures `SpotWrapper` properly releases all resources associated to leasing (including keepalives). It also extends leasing APIs so that downstream code needs not tinker with `SpotWrapper` internals.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -996,13 +996,18 @@ class SpotWrapper:
 
     def getLease(self) -> typing.Optional[Lease]:
         """Get a lease for the robot and keep the lease alive automatically."""
-        if not self._lease:
-            if self._use_take_lease:
-                self._lease = self._lease_client.take()
-            else:
-                self._lease = self._lease_client.acquire()
-
+        if self._use_take_lease:
+            lease = self._lease_client.take()
+        else:
+            lease = self._lease_client.acquire()
+        have_new_lease = (self._lease is None and lease is not None) or (
+            str(lease.lease_proto) != str(self._lease.lease_proto)
+        )
+        if have_new_lease:
+            if self._lease_keepalive is not None:
+                self._lease_keepalive.shutdown()
             self._lease_keepalive = LeaseKeepAlive(self._lease_client)
+            self._lease = lease
         return self._lease
 
     def releaseLease(self) -> None:
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -944,6 +944,8 @@ class SpotWrapper:
         """Get keepalive for eStop"""
         self._estop_endpoint = EstopEndpoint(self._estop_client, SPOT_CLIENT_NAME, self._estop_timeout)
         self._estop_endpoint.force_simple_setup()  # Set this endpoint as the robot's sole estop.
+        if self._estop_keepalive:
+            self._estop_keepalive.shutdown()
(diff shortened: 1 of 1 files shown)


spot_wrapper/wrapper.py
```diff
@@ -975,21 +977,39 @@ def releaseEStop(self) -> None:
...
         """Get a lease for the robot and keep the lease alive automatically."""
-        if self._use_take_lease:
-            self._lease = self._lease_client.take()
-        else:
-            self._lease = self._lease_client.acquire()
+        if not self._lease:
+            if self._use_take_lease:
+                self._lease = self._lease_client.take()
+            else:
+                self._lease = self._lease_client.acquire()
```
jbarry-bdai, line 1003: So I think we probably still need to call `take` or `acquire` because leases can be overridden.  My guess is that as written here once something takes control from the driver, the driver won't be able to take it back.

jbarry-bdai (commented):
Thanks for looking into this!

We should do some on-robot testing:

- [x] Start driver and confirm lease is not taken on driver start
- [x] Move the robot using the driver and ensure lease is taken and robot moves
- [ ] Hijack the robot with the tablet and ensure the robot can be driven around without the driver trying to take the lease back
- [ ] Move the robot using the driver and ensure the lease is taken again (even though a keepalive exists)

spot_wrapper/wrapper.py
mhidalgo-bdai: Yeap, you are right. I misunderstood the `claim` implementation. Fixed in https://github.com/bdaiinstitute/spot_wrapper/pull/92/commits/4463731cba1330827f727799229c1d3c2f452b40.

khughes-bdai:
Will work on testing those cases today @jbarry-bdai

tcappellari-bdai (dismissed):
whoops misread the above - ignore me

khughes-bdai:
On-robot testing with this seemed successful. I tried running `example_walk_forward`, `example_walk_with_velocity`, and  `example_ik_solutions` in `spot_utilities/examples`. One thing to note was that there were occasionally some spew error messages from the spot driver about the lease when I ran `example_walk_with_velocity` as well as `ros2 run spot_utilities spot_teleop.py Kepler`:
```
[spot_ros2-1] [wrapper.py:1062] Unable to execute robot command: bosdyn.api.RobotCommandResponse (LeaseUseError): 
[spot_ros2-1] [wrapper.py:1062] Unable to execute robot command: bosdyn.api.RobotCommandResponse (LeaseUseError): 
[spot_ros2-1] [wrapper.py:1062] Unable to execute robot command: bosdyn.api.RobotCommandResponse (LeaseUseError): 
[spot_ros2-1] [wrapper.py:1062] Unable to execute robot command: bosdyn.api.RobotCommandResponse (LeaseUseError): 
```
Is there a way we can reduce the spew of these messages?

I also tried taking control with the tablet while a longer example was actively running (`example_ik_requests`). In this case the example kept re-taking control from the tablet -- this is the desired behavior, right?

amessing-bdai (changes requested):
@khughes-bdai 
`Unable to execute robot command: bosdyn.api.RobotCommandResponse (LeaseUseError): ` implies there is an error?

khughes-bdai:
@amessing-bdai those errors existed before this, will make a ticket. Sorry for the confusion.

khughes-bdai (approved):
LGTM

amessing-bdai (approved):

jbarry-bdai (approved):

PR #103, opened 2024-04-12
Use secure gRPC channels for `spot_wrapper.testing` mocks

mhidalgo-bdai:
Pretty much what the title says. This patch drops `bosdyn` monkeypatching in favor of self-signed SSL certificates to allow for  secure gRPC channels during `spot_wrapper.testing` driven tests.


amessing-bdai (approved):

mhidalgo-bdai:
I will hold this off until I have a matching PR for https://github.com/bdaiinstitute/spot_ros2 (in the works).

mhidalgo-bdai:
FTR this certificates and keys are dummy ones. I don't know if we have a policy w.r.t. these nonetheless.

bhung-bdai (approved):
LGTM. Out of curiosity, does this represent the actual method that is used by Spot during authentication or is it just the next best thing?

mhidalgo-bdai:
> does this represent the actual method that is used by Spot during authentication

This allows us to use the Spot SDK as-is, so if it's not exactly equivalent to what the real robot does, I'd imagine it's rather close.

mhidalgo-bdai:
Testing went well here and in https://github.com/bdaiinstitute/spot_ros2/pull/347. Merging!

PR #104, opened 2024-04-26
Added possibility to directly control the arm using an ArmVelocityCommand

MGBla:
## Change Overview

Added the ability to set a Velocity command for the arm. This enables a user to have direct translational and rotational  control of the TCP movement using a controller or any other input device. The corresponding [PR](https://github.com/bdaiinstitute/spot_ros2/pull/366) on the driver Repo provides a topic interface to send the corresponding movement command. 
The angular and Cartesian velocity were thoroughly tested during operation on our spot. For the cylindrical velocity I performed some basic tests but didn't evaluate over hours as the other two input methods since I don't really use them in this reference frame.

## Testing Done
    - [x] tested that these new functionality work on robot

### DIFF
--- a/spot_wrapper/spot_arm.py
+++ b/spot_wrapper/spot_arm.py
@@ -500,6 +500,50 @@ class SpotArm:
 
         return True, "Moved arm successfully"
 
+    def handle_arm_velocity(
+        self, arm_velocity_command: arm_command_pb2.ArmVelocityCommand.Request, cmd_duration: float = 0.15
+    ) -> typing.Tuple[bool, str]:
+        """
+        Set the velocity of the arm TCP
+
+        Args:
+            arm_velocity_command: Protobuf message to set the arm velocity
+            cmd_duration: (optional) Time-to-live for the command in seconds.
+
+        Returns:
+            Boolean success, string message
+        """
+
+        try:
+            success, msg = self.ensure_arm_power_and_stand()
+            if not success:
+                self._logger.info(msg)
+                return False, msg
+            else:
+                end_time = self._robot.time_sync.robot_timestamp_from_local_secs(time.time() + cmd_duration)
+
+                arm_velocity_command2 = arm_command_pb2.ArmVelocityCommand.Request(
+                    cylindrical_velocity=arm_velocity_command.cylindrical_velocity,
+                    angular_velocity_of_hand_rt_odom_in_hand=arm_velocity_command.angular_velocity_of_hand_rt_odom_in_hand,
+                    cartesian_velocity=arm_velocity_command.cartesian_velocity,
+                    end_time=end_time,
+                )
+
+                robot_command = robot_command_pb2.RobotCommand()
+                robot_command.synchronized_command.arm_command.arm_velocity_command.CopyFrom(arm_velocity_command2)
+
+                self._robot_command_client.robot_command(
+                    command=robot_command, end_time_secs=time.time() + cmd_duration
+                )
+
+        except Exception as e:
+            print(e)
+            return (
+                False,
+                f"An error occured while trying to move arm\n Exception: {e}",
+            )
+        return True, "Moved arm successfully"
+
     @staticmethod
     def block_until_gripper_command_completes(
         robot_command_client: RobotCommandClient,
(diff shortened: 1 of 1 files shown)


khughes-bdai (approved):
LGTM

spot_wrapper/spot_arm.py
```diff
@@ -500,6 +500,50 @@ def hand_pose(self, data) -> typing.Tuple[bool, str]:
...
+                return False, msg
+            else:
+                end_time = self._robot.time_sync.robot_timestamp_from_local_secs(time.time() + cmd_duration)
+
+                arm_velocity_command2 = arm_command_pb2.ArmVelocityCommand.Request(
+                    cylindrical_velocity=arm_velocity_command.cylindrical_velocity,
+                    angular_velocity_of_hand_rt_odom_in_hand=arm_velocity_command.angular_velocity_of_hand_rt_odom_in_hand,
+                    cartesian_velocity=arm_velocity_command.cartesian_velocity,
+                    end_time=end_time,
+                )
```
tcappellari-bdai, line 530: Why do you copy the request to `arm_velocity_command2` here?

spot_wrapper/spot_arm.py
```diff
@@ -500,6 +500,50 @@ def hand_pose(self, data) -> typing.Tuple[bool, str]:
...
+
+                robot_command = robot_command_pb2.RobotCommand()
+                robot_command.synchronized_command.arm_command.arm_velocity_command.CopyFrom(arm_velocity_command2)
+
+                self._robot_command_client.robot_command(
+                    command=robot_command, end_time_secs=time.time() + cmd_duration
+                )
```
tcappellari-bdai, line 537: It would make more sense to use the `wait_for_arm_command_to_complete` method we already implemented rather than an arbitrary duration
```suggestion
                cmd_id = self._robot_command_client.robot_command(robot_command)
                self.wait_for_arm_command_to_complete(cmd_id)
```

spot_wrapper/spot_arm.py
tcappellari-bdai: oh now I see that you modified the end time.
Could you not do something like this?
```Python
new_end_time = . . .
arm_velocity_command.end_time = new_end_time
```

tcappellari-bdai (commented):
There are a few things I would suggest editing/want some clarification on, but overall looks fine on the wrapper side of things

spot_wrapper/spot_arm.py
amessing-bdai: @tcappellari-bdai I think those are 2 different behaviors as I don't think `end_time_secs` makes it a blocking call. Also if we are calling this inside a `service` in `spot_ros2` it shouldn't be a block call.

MGBla: Yes your right @amessing-bdai . And especially since it rather should be a topic subscription I don't want anything blocking in there.

spot_wrapper/spot_arm.py
```diff
@@ -500,6 +500,50 @@ def hand_pose(self, data) -> typing.Tuple[bool, str]:
...
+
+        except Exception as e:
+            print(e)
```
khughes-bdai, line 540:
```suggestion
```
can we remove this line?

spot_wrapper/spot_arm.py
khughes-bdai, line 530:
```suggestion
                arm_velocity_command2 = arm_command_pb2.ArmVelocityCommand.Request(
                    cylindrical_velocity=arm_velocity_command.cylindrical_velocity,
                    angular_velocity_of_hand_rt_odom_in_hand=arm_velocity_command.angular_velocity_of_hand_rt_odom_in_hand,
                    cartesian_velocity=arm_velocity_command.cartesian_velocity,
                    maximum_acceleration = arm_velocity_command.maximum_acceleration,
                    end_time=end_time,
                )
```

khughes-bdai (changes requested):
Just a few comments, but overall looks good

robodreamer:
Thanks for the update. Potentially, this provides another way to teleop the arm, since the current teleop uses a position command, but we could use the velocity command directly. The behavior around singularity and joint limits might be something to check once this PR gets merged.

khughes-bdai (approved):
LGTM

PR #109, opened 2024-05-15
[Sw-949] add mission client in spot wrapper

tcappellari-bdai:
Adds mission services to spot wrapper
Made `spot_mission_wrapper.py` and added service to `MockDirectoryService`

Previously, CI got stuck because the service was missing in `MockDirectoryService`. I ran the tests locally and they all pass now


khughes-bdai (approved):
LGTM

PR #111, opened 2024-05-17
uses claw_gripper_open_fraction_command for gripper_angle_open

### LINKED ISSUE #110: `gripper_angle_open` sets incorrect angle
when using `gripper_angle_open` as defined in [`spot_arm.py`](https://github.com/bdaiinstitute/spot_wrapper/blob/main/spot_wrapper/spot_arm.py#L419) the angle of the gripper appears to be incorrect. Here are two images where I tried setting the angle to 90 degrees. In the first image i used `gripper_angle_open`, in the second one i used the `gripper_open_fraction_command`. 

image

image

For the fraction command, the angle looks correct (used a fraction f 1.0), while for the angle the gripper has a smaller angle than it should have. When using the 'open_gripper' command, the angle is the same as using the fraction command. Same occurs when using 45 degrees. The angle command also sometimes results in jittering, but I don't know if that is related.

Where do the limits used on [line 422](https://github.com/bdaiinstitute/spot_wrapper/blob/main/spot_wrapper/spot_arm.py#L422) come from? I couldn't find them in the urdf files.

taDachs:
This is a hotfix that resolves #110. It was tested on a real robot.

### DIFF
--- a/spot_wrapper/spot_arm.py
+++ b/spot_wrapper/spot_arm.py
@@ -417,7 +417,7 @@ class SpotArm:
                 self._logger.info(msg)
                 return False, msg
             else:
-                command = RobotCommandBuilder.claw_gripper_open_fraction_command(angle / 90.0)
+                command = RobotCommandBuilder.claw_gripper_open_fraction_command(gripper_ang / 90.0)
 
                 # Command issue with RobotCommandClient
                 cmd_id = self._robot_command_client.robot_command(command)
--- a/spot_wrapper/spot_arm.py
+++ b/spot_wrapper/spot_arm.py
@@ -417,12 +417,7 @@ class SpotArm:
                 self._logger.info(msg)
                 return False, msg
             else:
-                # The open angle command does not take degrees but the limits
-                # defined in the urdf, that is why we have to interpolate
-                closed = 0.349066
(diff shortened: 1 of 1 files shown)


spot_wrapper/spot_arm.py
```diff
@@ -417,12 +417,7 @@ def gripper_angle_open(self, gripper_ang: float) -> typing.Tuple[bool, str]:
                 self._logger.info(msg)
                 return False, msg
             else:
-                # The open angle command does not take degrees but the limits
-                # defined in the urdf, that is why we have to interpolate
-                closed = 0.349066
-                opened = -1.396263
-                angle = gripper_ang / 90.0 * (opened - closed) + closed
-                command = RobotCommandBuilder.claw_gripper_open_angle_command(angle)
+                command = RobotCommandBuilder.claw_gripper_open_fraction_command(angle / 90.0)
```
khughes-bdai, line 420:
```suggestion
                command = RobotCommandBuilder.claw_gripper_open_fraction_command(gripper_ang / 90.0)
```

khughes-bdai (changes requested):
Right now this doesn't run because `angle` isn't defined, but once that's changed it looks good to me! Thanks

PR #116, opened 2024-06-17
Add functionality to set the grasp and carry overrides

jbarry-bdai:
This adds a function that lets us override grasp state (whether or not the robot thinks it's carrying something) and/or carry state (what the robot will do on stow) through the spot wrapper.  A companion request will be open in spot_ros2 to wrap this in a ros2 service.

**Testing Done**
All testing done by calling the ROS2 service from the command line and then checking the manipulation state topic:
- [x] Only grasp override set in request message
- [x] Only carry override set in the request message
- [x] Grasp state override and carry state explicitly set to 0 in request message
- [x] Carry override and grasp override set in request message
- [x] Carry override set and grasp override set to NOT_HOLDING (correctly returns a False and an error message)

I also spot checked that the tablet Empty/Grasping icon changed appropriately and checked the hijack behavior to see the carry state effects.

### DIFF
--- a/spot_wrapper/spot_arm.py
+++ b/spot_wrapper/spot_arm.py
@@ -614,8 +614,11 @@ class SpotArm:
         except Exception as e:
             return False, f"An error occured while trying to grasp from pose {e}"
 
-    def override_grasp_or_carry(self, grasp_override: manipulation_api_pb2.ApiGraspOverride.Override,
-                                carry_override: ManipulatorState.CarryState) -> typing.Tuple[bool, str]:
+    def override_grasp_or_carry(
+        self,
+        grasp_override: manipulation_api_pb2.ApiGraspOverride.Override,
+        carry_override: ManipulatorState.CarryState,
+    ) -> typing.Tuple[bool, str]:
         """
         Override the robot's grasp and/or carry state.
 
--- a/spot_wrapper/spot_arm.py
+++ b/spot_wrapper/spot_arm.py
@@ -11,6 +11,7 @@ from bosdyn.api import (
     synchronized_command_pb2,
     trajectory_pb2,
 )
+from bosdyn.api.robot_state_pb2 import ManipulatorState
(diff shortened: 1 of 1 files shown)


spot_wrapper/spot_arm.py
```diff
@@ -612,3 +613,49 @@ def grasp_3d(self, frame: str, object_rt_frame: typing.List[float]) -> typing.Tu
                 return False, msg
         except Exception as e:
             return False, f"An error occured while trying to grasp from pose {e}"
+
+    def override_grasp_or_carry(self, grasp_override: manipulation_api_pb2.ApiGraspOverride.Override,
```
khughes-bdai, line 617: nit: maybe have these as `Optional`s so that you can pass in `None` if you don't care about overriding one of the states?

jbarry-bdai: You use the UNKNOWN (0) choice for that for both of them.

jbarry-bdai: I started with them as optional and then realized it was pretty annoying because I was converting from the protobuf types anyway...

khughes-bdai (approved):
LGTM

spot_wrapper/spot_arm.py
```diff
@@ -612,3 +613,52 @@ def grasp_3d(self, frame: str, object_rt_frame: typing.List[float]) -> typing.Tu
                 return False, msg
         except Exception as e:
             return False, f"An error occured while trying to grasp from pose {e}"
+
+    def override_grasp_or_carry(
```
amessing-bdai, line 617: Is there a reason to bundle these or should this just be 2 functions?

jbarry-bdai:
@khughes-bdai had to do some updating... mind re-approving the most recent commit?

mhidalgo-bdai (approved):
LGTM

PR #120, opened 2024-06-25
Extend response header autocompletion support

mhidalgo-bdai:
Follow-up to #81. This patch extends the logic for response header autocompletion to gRPC services that are not unary in both request and response.

### DIFF
--- a/spot_wrapper/testing/helpers.py
+++ b/spot_wrapper/testing/helpers.py
@@ -3,12 +3,10 @@
 import functools
 import typing
 
-import grpc
-from bosdyn.api.header_pb2 import CommonError
 from bosdyn.api.lease_pb2 import ResourceTree
 
 
-class GeneralizedDecorator:
+class ThinDecorator:
     __name__ = __qualname__ = __doc__ = ""
 
     __annotations__ = {}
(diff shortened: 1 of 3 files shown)


spot_wrapper/testing/helpers.py
```diff
@@ -3,20 +3,18 @@
 import functools
 import typing
 
-import grpc
-from bosdyn.api.header_pb2 import CommonError
 from bosdyn.api.lease_pb2 import ResourceTree
 
 
-class GeneralizedDecorator:
+class ThinDecorator:
```
amessing-bdai, line 9: Any chance we can get a more descriptive name?

mhidalgo-bdai: I'm not always great with names. `TransparentDecorator`? `ForwardingDecorator` is almost tautological.

amessing-bdai: Naming might be the hardest thinking in programming, but I can't tell you what a `ThinDecorator` does just by reading the name. Maybe just add a usage comment?

mhidalgo-bdai: How about `ForwardingWrapper`? https://github.com/bdaiinstitute/spot_wrapper/pull/120/commits/6901afa781bdb47f1f739a5b0d867bf9a7cd6651

amessing-bdai (approved):

PR #124, opened 2024-07-10
Avoid using port 0 for `SpotCamWrapper`

mhidalgo-bdai:
Follow-up to #81. `SpotCamWrapper` should never try to use an arbitrary, random port.


tcappellari-bdai (approved):
lgtm

khughes-bdai (approved):

PR #125, opened 2024-07-15
[SW-822] Spot Driver without gripper

tcappellari-bdai:
## Change Overview

Added a parameter "gripperless" to indicate if the gripper is attached to the arm or not. If the arm is present but gripper is not, it will skip making the gripper and hand camera services.

Goes with https://github.com/bdaiinstitute/spot_ros2/pull/430

## Testing Done
Tested on a Spot with a special gripperless firmware

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -343,6 +343,7 @@ class SpotWrapper:
         rgb_cameras: bool = True,
         payload_credentials_file: str = None,
         cert_resource_glob: typing.Optional[str] = None,
+        gripperless: bool = False,
     ) -> None:
         """
         Args:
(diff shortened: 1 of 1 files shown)


spot_wrapper/wrapper.py
```diff
@@ -443,7 +445,7 @@ def __init__(
                 self._spot_check_client = self._robot.ensure_client(SpotCheckClient.default_service_name)
                 self._mission_client = self._robot.ensure_client(MissionClient.default_service_name)
                 self._license_client = self._robot.ensure_client(LicenseClient.default_service_name)
-                if self._robot.has_arm():
+                if self._robot.has_arm() and not self.gripperless:
```
amessing-bdai, line 448: You should swap these conditions. `has_arm` involves a grpc call.

tcappellari-bdai: good point

amessing-bdai (approved):

bhung-bdai (approved):

Issue #47, opened 2023-06-30
Restructure `AsyncIdle`


heuristicus:
Currently `AsyncIdle` has a mixture of uses and does quite a few different things depending on the state of various flags.

I don't really know the best way to do this, but perhaps starting by splitting the function up into separate functions handling each of the commands makes sense.

The variables that are now in `RobotState` are set by its `_start_query` method. But I think some of these variables are probably not correctly tracking the actual state of the robot. Ideally this information would come from the SDK in some way but I'm not sure that's possible.

It also sends the stand command repeatedly if no command is running. I'm not sure this is really necessary.

PR #130, opened 2024-08-16
Handle cases when mobility command feedback does not contain a specific command status

Imaniac230:
Hi,

I have stumbled upon a bit obscure error message from the [ROS2 driver](https://github.com/bdaiinstitute/spot_ros2)  when attempting to send `sit` commands to the robot while it was also actively receiving velocity `movement` commands. This might also be related to:
* https://github.com/bdaiinstitute/spot_wrapper/issues/47
* https://github.com/bdaiinstitute/spot_ros2/issues/260

It makes sense that the `sit` commands will always be overridden by the `movement` commands and therefore the robot will refuse to sit down. However, the ROS2 driver responds to this by a outputting an error message, ex.: `[spot_ros2-5] [wrapper.py:236] Error when getting robot command feedback: bosdyn.api.RobotCommandFeedbackResponse (InternalServerError): Command not found, id: 62532`, which was quite confusing.

I have found the root cause of the error in this parsing logic: https://github.com/bdaiinstitute/spot_wrapper/blob/99ec4a61d9949e904e79474e0b832a072076d03e/spot_wrapper/wrapper.py#L223-L237
, specifically this status parsing:
https://github.com/bdaiinstitute/spot_wrapper/blob/99ec4a61d9949e904e79474e0b832a072076d03e/spot_wrapper/wrapper.py#L227-L230

The normal feedback response looks like this:
```
feedback {
  synchronized_feedback {
    mobility_command_feedback {
      sit_feedback {
        status: STATUS_IS_SITTING
      }
      status: STATUS_PROCESSING
    }
  }
}
```
However, when the command is overridden, the feedback response looks like this:
```
feedback {
  synchronized_feedback {
    mobility_command_feedback {
      status: STATUS_COMMAND_OVERRIDDEN
    }
  }
}
```
and this is where the the logic throws an exception because the `sit_feedback` section does not exist in the message at all. The same also applies to the `stand` command. I think the same might also apply to the `trajectory` command, but I have not tested that one.

I have added a top-level check for the value of `mobility_command_feedback.status` before any further parsing is performed. If this status is not currently `STATUS_PROCESSING` then the command is no longer valid and its `last_*_command` variable is reset together with a warning message.

This does not modify the actual robot behavior in any way, but avoids the obscure error message and instead handles the cases explicitly. I'm not sure if there are any larger refactors for the `AsyncIdle` class being currently in the works relating to https://github.com/bdaiinstitute/spot_wrapper/issues/47 and so if this small proposal has any value currently, but I am putting it out here.

I've also added a final check at the end, which might relate to the issue here https://github.com/bdaiinstitute/spot_ros2/issues/260. If the robot is in a `moving` state, the `sitting` and `standing` states will be set accordingly. This would make sense since the `movement` commands currently always force the robot to stand up a will override any `sit` and `stand` results. But, this may not be an optimal solution if the behavior changes in the future.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -206,32 +206,43 @@ class AsyncIdle(AsyncPeriodicQuery):
         if self._spot_wrapper.last_stand_command is not None:
             try:
                 response = self._client.robot_command_feedback(self._spot_wrapper.last_stand_command)
-                status = response.feedback.synchronized_feedback.mobility_command_feedback.stand_feedback.status
-                self._spot_wrapper.is_sitting = False
-                if status == basic_command_pb2.StandCommand.Feedback.STATUS_IS_STANDING:
-                    self._spot_wrapper.is_standing = True
-                    self._spot_wrapper.last_stand_command = None
-                elif status == basic_command_pb2.StandCommand.Feedback.STATUS_IN_PROGRESS:
-                    self._spot_wrapper.is_standing = False
+                command_feedback = response.feedback.synchronized_feedback.mobility_command_feedback
+                if command_feedback.status == basic_command_pb2.RobotCommandFeedbackStatus.STATUS_PROCESSING:
+                    self._spot_wrapper.is_sitting = False
+                    stand_status = command_feedback.stand_feedback.status
+                    if stand_status == basic_command_pb2.StandCommand.Feedback.STATUS_IS_STANDING:
+                        self._spot_wrapper.is_standing = True
+                        self._spot_wrapper.last_stand_command = None
+                    elif stand_status == basic_command_pb2.StandCommand.Feedback.STATUS_IN_PROGRESS:
+                        self._spot_wrapper.is_standing = False
+                    else:
+                        self._logger.warning("Stand command in unknown state")
+                        self._spot_wrapper.is_standing = False
                 else:
-                    self._logger.warning("Stand command in unknown state")
-                    self._spot_wrapper.is_standing = False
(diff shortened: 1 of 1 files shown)


mhidalgo-bdai (approved):
Fix looks appropriate though a regression test would be nice.

Imaniac230:
I see that the linter tests are failing. I'll get it fixed as soon as I get back to it. Sorry for that, I should have ran the linter checks beforehand.

khughes-bdai (approved):
Looks good to me once the linting is fixed. I just tested on robot and verified that it does resolve https://github.com/bdaiinstitute/spot_ros2/issues/260. Thanks!

Imaniac230:
Looks like the uploading to Coveralls stage is failing with a missing repo token:
```
Missing Repo Token. Set using `-r <token>` or `COVERALLS_REPO_TOKEN=<token>`
```
I'm not sure if there is anything I can try to fix that without actually triggering multiple CI runs.

mhidalgo-bdai:
@amessing-bdai any clue what may be going on here? Was the token removed? Did it expire?

amessing-bdai:
> @amessing-bdai any clue what may be going on here? Was the token removed? Did it expire?

No, it wasn't removed and shouldn't have expired (there isn't a timeout on it that I can see)

khughes-bdai:
I think that the token issue is because this PR is from outside of the `bdai` organization and doesn't have the same permissions. I have seen it happen before on other external PRs. I am not sure why the tests are happening on 3 different versions of Python, but  the actual tests are passing, it is just the coveralls step that is failing.

tcappellari-bdai (approved):

PR #132, opened 2024-08-21
Implement tests for #130.

Imaniac230:
Hi, I am creating a followup to https://github.com/bdaiinstitute/spot_wrapper/pull/130 with added tests for the status handling and state changes. I wanted to test as many cases as possible, so I hope it's not too bloated.

The tests check the following new scenarios (these checks should be passing with the recent changes and would not be passing before the changes):
1. When the mobility command feedback status is not `STATUS_PROCESSING`, whether the `last_sit_command`/`last_stand_command` is correctly reset and the `is_sitting`/`is_standing` state remains unmodified.
2. When a velocity movement command is processed (trajectory commands are currently not included in the test), whether the `is_sitting` and `is_standing` states are reflecting the change.

The tests also check the already existing behavior when handling a valid sit/stand feedback status (`STATUS_IN_PROGRESS`, `STATUS_IS_STANDING`, `STATUS_IS_SITTING`).

### DIFF
--- a/spot_wrapper/tests/test_wrapper.py
+++ b/spot_wrapper/tests/test_wrapper.py
@@ -141,7 +141,7 @@ def test_mobility_command_feedback(simple_spot: SpotFixture, simple_spot_wrapper
     simple_spot.api.RobotCommand.future.returns(response).forever()
 
     def update_with_feedback_response(r: RobotCommandFeedbackResponse) -> None:
-        simple_spot.api.RobotCommandFeedback.future.returns(r).repeatedly(1)
+        simple_spot.api.RobotCommandFeedback.future.returns(r)
         simple_spot_wrapper.updateTasks()
         time.sleep(0.1)
 
--- a/spot_wrapper/tests/test_wrapper.py
+++ b/spot_wrapper/tests/test_wrapper.py
@@ -1,10 +1,12 @@
 # Copyright (c) 2023 Boston Dynamics AI Institute LLC. See LICENSE file for more info.
 
 import logging
-from typing import Iterator
+import time
+from typing import Iterator, List
 
 import grpc
 import pytest
+from bosdyn.api.basic_command_pb2 import RobotCommandFeedbackStatus, SitCommand, StandCommand
 from bosdyn.api.manipulation_api_pb2 import (
     ManipulationApiRequest,
     ManipulationApiResponse,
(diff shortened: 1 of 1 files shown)


spot_wrapper/tests/test_wrapper.py
```diff
@@ -120,3 +122,144 @@ def test_spot_wrapper_commands(simple_spot: SpotFixture, simple_spot_wrapper: Sp
...
+    simple_spot_wrapper.velocity_cmd(0, 0, 0)
+    assert simple_spot_wrapper.last_velocity_command_time is not None
+    simple_spot_wrapper.updateTasks()
+    time.sleep(0.1)
+
+    assert simple_spot_wrapper.last_sit_command is None
+    assert simple_spot_wrapper.last_stand_command is None
+    assert not simple_spot_wrapper.is_sitting
+    assert simple_spot_wrapper.is_standing
+    assert simple_spot_wrapper.is_moving
```
khughes-bdai, line 265: Question: shouldn't a velocity command of 0 not stand up+move the robot in practice?

Imaniac230: From what I have tested practically with the robot, once it starts receiving any velocity commands (zero included) it will always stand up and override any following sit/stand commands.

Logically, I think it does make sense (in relation to the intended behavior discussed here https://github.com/bdaiinstitute/spot_ros2/issues/175), because any valid velocity command implies that the robot should to be ready to move, and a zero velocity simply means standing in place.

khughes-bdai: Ok, makes sense!

khughes-bdai (approved):
LGTM -- tested locally and tests are all passing. Thanks!

spot_wrapper/tests/test_wrapper.py
```diff
@@ -120,3 +122,144 @@ def test_spot_wrapper_commands(simple_spot: SpotFixture, simple_spot_wrapper: Sp
...
+
+    # bosdyn.api.RobotCommandService/RobotCommand implementation is mocked
+    # (as spot_wrapper.testing.mocks.MockSpot is automatically specified),
+    # so here we provide the same response for every future command.
+    response = RobotCommandResponse()
+    response.status = RobotCommandResponse.Status.STATUS_OK
+    simple_spot.api.RobotCommand.future.returns(response).forever()
+
+    def update_with_feedback_response(r: RobotCommandFeedbackResponse) -> None:
+        simple_spot.api.RobotCommandFeedback.future.returns(r).repeatedly(1)
```
mhidalgo-bdai, line 144: @Imaniac230 fyi: 
```suggestion
        simple_spot.api.RobotCommandFeedback.future.returns(r)
```
is equivalent to what you have.

mhidalgo-bdai (approved):
LGTM2

spot_wrapper/tests/test_wrapper.py
Imaniac230: That never occurred to me, thanks.

khughes-bdai (approved):

tcappellari-bdai (approved):

PR #138, opened 2024-09-04
Calibration fixes: Supporting OpenCV prior to and after breaking changes in 4.7

glvov-bdai:
In OpenCV 4.7.0, charuco was moved from the contrib part of opencv to the core part of opencv (https://github.com/opencv/opencv/pull/22986/files#diff-09508dfcb283f5b8f2e58538584a58ce28d369f8c7e23705f004f7d15e6f40e9). Previously, the code assumed an OpenCV version prior to 4.7.0.

This adds an explicit check for the opencv version prior to prior to many charuco operations, so that the correct charuco method can be called for opencv>=4.7 despite the breaking change in 4.7

More information about OpenCV changes:
https://github.com/opencv/opencv/issues/26126

This also adds an explicit dependency on opencv. Previously, this dependency was likely implicitly included if you use the spot_ros2 package (due to dependency on things like cv bridge), but this not cover the use case of using spot_wrapper completely independently of ROS2.

It is better to check OpenCV version, and adjust behavior accordingly, then just depending on opencv-contrib-python, as
depending on contrib can break existing versions of OpenCV on your system when the requirements.txt is installed

Testing:

Tested on physical robot. For code changes with OpenCV, verified the results in prior to 4.7 and after:
https://github.com/opencv/opencv/issues/26126

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,6 +7,7 @@ bosdyn-mission==4.0.2
 grpcio==1.59.3
 image==1.5.33
 inflection==0.5.1
+opencv-python>=4.5.4
 open3d==0.18.0 
 protobuf==4.22.1
 pytest==7.3.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -13,4 +13,5 @@ protobuf==4.22.1
 pytest==7.3.1
 pytest-cov==4.1.0
 pytest-xdist==3.5.0
+pyyaml>=6.0
 setuptools==59.6.0
--- a/spot_wrapper/calibration/README.md
+++ b/spot_wrapper/calibration/README.md
@@ -168,8 +168,12 @@ python3 calibrate_spot_hand_camera_cli.py --ip <IP> -u user -pw <SECRET> --data_
 ## Improving Calibration Quality
 If you find that the calibration quality isn't high enough, try a longer calibration
 with a wider variety of viewpoints (decrease the step size, increase the bounds). 
-The default calibration viewpoint parameters are meant to facilitate a quick  calibration
+The default calibration viewpoint parameters are meant to facilitate a quick calibration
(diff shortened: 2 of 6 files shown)


glvov-bdai:
* **#138** <a href="https://app.graphite.dev/github/pr/bdaiinstitute/spot_wrapper/138?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈
* `main`

This stack of pull requests is managed by Graphite. <a href="https://stacking.dev/?utm_source=stack-comment">Learn more about stacking.</a>
<h2></h2>

 Join @glvov-bdai and the rest of your teammates on <a href="https://graphite.dev?utm-source=stack-comment">Graphite <b>Graphite</b></a>

requirements.txt
```diff
@@ -7,9 +7,11 @@ bosdyn-mission==4.0.2
 grpcio==1.59.3
 image==1.5.33
 inflection==0.5.1
+opencv-python>=4.5.4
```
amessing-bdai, line 10: Is there a way to not add this here? When you depend on opencv via rosdep in ros 1 or ros 2, it is pulled in via apt. We have had problems in the past with conflicting versions of opencv between apt and pip.

amessing-bdai: @jbarry-bdai @khughes-bdai @tcappellari-bdai Thoughts? It should be a dependency for this to be a stand alone package, but then `spot_ros2` needs to uninstall it similar to what we do elsewhere which isn't great.

glvov-bdai: I would maybe suggest just splitting up the requirements.txt between `spot_ros2` and `spot_wrapper`, and then using that to manage dependencies for both independently

tcappellari-bdai: ya i like gary's suggestion of splitting them up (assuming that wouldn't break everything lol)

tcappellari-bdai (approved):

PR #139, opened 2024-09-06
Arm joint commands wrapper function

llee-bdai:
Adds a simple function to the wrapper to allow us to simply send arm joint angles to the robot.

Tested with the Maple `action_stack_spot` sending joint angles.


spot_wrapper/wrapper.py
```diff
@@ -1302,6 +1304,20 @@ def velocity_cmd(
...
+        self, sh0: float, sh1: float, el0: float, el1: float, wr0: float, wr1: float
+    ) -> typing.Tuple[bool, str]:
+        traj_point = RobotCommandBuilder.create_arm_joint_trajectory_point(sh0, sh1, el0, el1, wr0, wr1)
+        arm_joint_traj = arm_command_pb2.ArmJointTrajectory(points=[traj_point])
+
+        joint_move_command = arm_command_pb2.ArmJointMoveCommand.Request(trajectory=arm_joint_traj)
+        arm_command = arm_command_pb2.ArmCommand.Request(arm_joint_move_command=joint_move_command)
+        sync_arm_command = synchronized_command_pb2.SynchronizedCommand.Request(arm_command=arm_command)
+        arm_sync_robot_cmd = robot_command_pb2.RobotCommand(synchronized_command=sync_arm_command)
+        robot_command = RobotCommandBuilder.build_synchro_command(arm_sync_robot_cmd)
```
amessing-bdai, line 1317:
```suggestion
        robot_command = RobotCommandBuilder.arm_joint_command(sh0=sh0, sh1=sh1, el0=el0, el1=el1, wr0=wr0, wr1=wr1)
```

Is there a reason not to just use this?

spot_wrapper/wrapper.py
```diff
@@ -1302,6 +1304,20 @@ def velocity_cmd(
         self.last_velocity_command_time = end_time
         return response[0], response[1]
 
+    def arm_joint_cmd(
+        self, sh0: float, sh1: float, el0: float, el1: float, wr0: float, wr1: float
```
amessing-bdai, line 1308:
```suggestion
        self, *, sh0: float, sh1: float, el0: float, el1: float, wr0: float, wr1: float
```

When you have a bunch of arguments with the same type that can easily get swapped, I would recommend keyword guarding them.

spot_wrapper/wrapper.py
llee-bdai: because `arm_joint_command` is a function we have defined in `SpotBasic` which goes through these exact same steps, but the point of adding this to the wrapper directly is because i'm trying to expose this without having to use `SpotBasic` for functional reasons....

spot_wrapper/wrapper.py
llee-bdai: ..... i'll be honest, not sure why we need the `*` input here...
also, I just tested it and it is keyword guarded isn't it?

llee-bdai: lol nevermind. Updated

spot_wrapper/wrapper.py
amessing-bdai: https://github.com/boston-dynamics/spot-sdk/blob/master/python/bosdyn-client/src/bosdyn/client/robot_command.py#L1422

We do have a function named that in `SpotBasic`, but I am suggesting using the one from the `RobotCommandBuilder` that comes from the `spot-sdk`

llee-bdai: Oh lol.... I missed that. Let me try that one on Monday on robot. It does the same thing it looks like as I'm doing above - which the only reason I'm doing above is because that's how the `arm_joint_move` example did it here https://github.com/boston-dynamics/spot-sdk/blob/master/python/examples/arm_joint_move/arm_joint_move.py

llee-bdai:
### Merge activity

* **Sep 10, 11:45 AM EDT**: @llee-bdai started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/bdaiinstitute/spot_wrapper/139).

amessing-bdai (approved):

tcappellari-bdai (approved):

PR #145, opened 2024-10-07
[MAPLE-743] Update hand_pose function

llee-bdai:
The `hand_pose()` function as is relies on ROS 1 data types. This update makes it more generic without any ROS type dependency

### DIFF
--- a/spot_wrapper/spot_arm.py
+++ b/spot_wrapper/spot_arm.py
@@ -457,7 +457,7 @@ class SpotArm:
         Args:
             x, y, z: float positions for the pose
             qx, qy, qz, qw: float quaternion for the pose
-            duration: secodns for the trajectory
+            duration: target duration of the trajectory, in seconds
             ref_frame: base frame for the pose. This needs to be something Spot knows about, ie "body" or "arm0.link_sh0"
             ensure_power_on_and_stand: bool for whether or not to ensure Spot is standing/powered on before executing
 
@@ -479,7 +479,7 @@ class SpotArm:
             # Build a position to move the arm to (in meters, relative to the body frame origin.)
             position = geometry_pb2.Vec3(x=x, y=y, z=z)
 
-            # # Rotation as a quaternion.
+            # Rotation as a quaternion.
             rotation = geometry_pb2.Quaternion(w=qw, x=qx, y=qy, z=qz)
 
             traj_duration = seconds_to_duration(duration)
--- a/spot_wrapper/spot_arm.py
+++ b/spot_wrapper/spot_arm.py
@@ -455,7 +455,11 @@ class SpotArm:
         Set the pose of the hand
 
         Args:
-            data:
+            x, y, z: float positions for the pose
+            qx, qy, qz, qw: float quaternion for the pose
+            duration: secodns for the trajectory
+            ref_frame: base frame for the pose. This needs to be something Spot knows about, ie "body" or "arm0.link_sh0"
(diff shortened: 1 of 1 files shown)


spot_wrapper/spot_arm.py
```diff
@@ -437,63 +437,73 @@ def gripper_angle_open(self, gripper_ang: float, ensure_power_on_and_stand: bool
...
+        ensure_power_on_and_stand: bool = True,
+    ) -> typing.Tuple[bool, str]:
         """
         Set the pose of the hand
 
         Args:
-            data:
+            x, y, z: float positions for the pose
+            qx, qy, qz, qw: float quaternion for the pose
+            duration: secodns for the trajectory
```
scastro-bdai, line 460:
```suggestion
            duration: Duration of the trajectory, in seconds

File: spot_wrapper/spot_arm.py [hunk 2]
```diff
@@ -437,63 +437,73 @@ def gripper_angle_open(self, gripper_ang: float, ensure_power_on_and_stand: bool
...
         try:
-            success, msg = self.ensure_arm_power_and_stand()
-            if not success:
-                self._logger.info(msg)
-                return False, msg
+            if ensure_power_on_and_stand:
+                success, msg = self.ensure_arm_power_and_stand()
+                if not success:
+                    self._logger.info(msg)
+                    return False, msg
```
scastro-bdai, line 472: This is not your doing, but it makes me anxious to see this much code all wrapped up in the same `try` block...

spot_wrapper/spot_arm.py
```diff
@@ -437,63 +437,73 @@ def gripper_angle_open(self, gripper_ang: float, ensure_power_on_and_stand: bool
...
-                    y=pose_point.orientation.y,
-                    z=pose_point.orientation.z,
-                )
+            # Move the arm to a spot in front of the robot given a pose for the gripper.
+            # Build a position to move the arm to (in meters, relative to the body frame origin.)
+            position = geometry_pb2.Vec3(x=x, y=y, z=z)
 
-                seconds = data.duration
-                duration = seconds_to_duration(seconds)
+            # # Rotation as a quaternion.
```
scastro-bdai, line 482: nit: remove double comment markers

```suggestion
            # Rotation as a quaternion.

**Review by scastro-bdai — APPROVED** [review]

File: spot_wrapper/spot_arm.py (hunk 2 above)
Comment (llee-bdai) [reply spot_wrapper/spot_arm.py:472]: I agree, but basically every single function in this module is in a `try... except` to make sure it's not a blocker and always returns I think

Comment (amessing-bdai) [reply spot_wrapper/spot_arm.py:472]: Agree as well. Especially since its a generic exception that is caught versus a specific one. But since it was already here I think it is ok for now. There is some discuss about some larger changes that will improve much of this.

File: spot_wrapper/spot_arm.py [hunk 4]
```diff
@@ -437,63 +437,73 @@ def gripper_angle_open(self, gripper_ang: float, ensure_power_on_and_stand: bool
...
+    ) -> typing.Tuple[bool, str]:
         """
         Set the pose of the hand
 
         Args:
-            data:
+            x, y, z: float positions for the pose
+            qx, qy, qz, qw: float quaternion for the pose
+            duration: target duration of the trajectory, in seconds
+            ref_frame: base frame for the pose. This needs to be something Spot knows about, ie "body" or "arm0.link_sh0"
```
amessing-bdai, line 461: CI is mad about how long this line is

amessing-bdai (approved):

PR #146, opened 2024-10-16
[MAPLE-743] Add non blocking to hand_pose and change default traj timing

llee-bdai:
Add in non blocking option to return before arm trajectory completes.

Also changes to the default trajectory timing to 0.0 (as fast as possible) unless specifically specified to be some duration.

### DIFF
--- a/spot_wrapper/spot_arm.py
+++ b/spot_wrapper/spot_arm.py
@@ -447,9 +447,10 @@ class SpotArm:
         qy: float,
         qz: float,
         qw: float,
-        duration: float = 0.125,
+        duration: float = 0.0,
         ref_frame: str = "body",
         ensure_power_on_and_stand: bool = True,
+        blocking: bool = True,
     ) -> typing.Tuple[bool, str]:
         """
         Set the pose of the hand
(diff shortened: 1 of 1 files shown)


spot_wrapper/spot_arm.py
```diff
@@ -447,9 +447,10 @@ def hand_pose(
         qy: float,
         qz: float,
         qw: float,
-        duration: float = 0.125,
+        duration: float = 0.0,
```
scastro-bdai, line 450: Will this cause an incompatibility with other users who may be relying on the nonzero default arg?

spot_wrapper/spot_arm.py
```diff
@@ -483,11 +484,14 @@ def hand_pose(
             # Rotation as a quaternion.
             rotation = geometry_pb2.Quaternion(w=qw, x=qx, y=qy, z=qz)
 
-            traj_duration = seconds_to_duration(duration)
-
             # Build the SE(3) pose of the desired hand position in the moving body frame.
             hand_pose = geometry_pb2.SE3Pose(position=position, rotation=rotation)
-            hand_pose_traj_point = trajectory_pb2.SE3TrajectoryPoint(pose=hand_pose, time_since_reference=traj_duration)
+            hand_pose_traj_point = trajectory_pb2.SE3TrajectoryPoint(pose=hand_pose)
+            if duration != 0.0:
```
scastro-bdai, line 490: Generally skeptical of floating point equality checks. If it were up to me, I'd maybe make `duration` be a `double | None`, so you can specifically set it to `None` or something.

spot_wrapper/spot_arm.py
llee-bdai: Seeing as I am the only user right now.... probably not?

spot_wrapper/spot_arm.py
llee-bdai: Updated to be optionally None. Thanks for catching that, I wasn't really going through everything with a fine toothcomb

scastro-bdai (approved):

llee-bdai:
### Merge activity

* **Oct 17, 10:13 AM EDT**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.dev/github/pr/bdaiinstitute/spot_wrapper/146).

PR #149, opened 2024-11-07
Default wrist_tform_tool to identity

llee-bdai:
Turns out `ArmCartesianCommand` defaults `wrist_tform_tool` to be some approximately 20cm offset https://dev.bostondynamics.com/protos/bosdyn/api/proto_reference#armcartesiancommand-request

This PR sets the transform to be identity so that when you send a hand pose, it's commanding the wrist to the pose, and not wrist plus some unknown defaulted offset.

### DIFF
--- a/spot_wrapper/spot_arm.py
+++ b/spot_wrapper/spot_arm.py
@@ -496,6 +496,7 @@ class SpotArm:
 
             arm_cartesian_command = arm_command_pb2.ArmCartesianCommand.Request(
                 root_frame_name=ref_frame,
+                wrist_tform_tool=geometry_pb2.SE3Pose(position=geometry_pb2.Vec3(x=0.0, y=0.0, z=0.0), rotation=geometry_pb2.Quaternion(w=1.0, x=0.0, y=0.0, z=0.0)),
                 pose_trajectory_in_task=hand_trajectory,
                 force_remain_near_current_joint_configuration=True,
             )
--- a/spot_wrapper/spot_arm.py
+++ b/spot_wrapper/spot_arm.py
@@ -496,7 +496,10 @@ class SpotArm:
 
             arm_cartesian_command = arm_command_pb2.ArmCartesianCommand.Request(
                 root_frame_name=ref_frame,
-                wrist_tform_tool=geometry_pb2.SE3Pose(position=geometry_pb2.Vec3(x=0.0, y=0.0, z=0.0), rotation=geometry_pb2.Quaternion(w=1.0, x=0.0, y=0.0, z=0.0)),
+                wrist_tform_tool=geometry_pb2.SE3Pose(
+                    position=geometry_pb2.Vec3(x=0.0, y=0.0, z=0.0),
+                    rotation=geometry_pb2.Quaternion(w=1.0, x=0.0, y=0.0, z=0.0),
+                ),
                 pose_trajectory_in_task=hand_trajectory,
                 force_remain_near_current_joint_configuration=True,
             )


spot_wrapper/spot_arm.py
```diff
@@ -496,6 +496,10 @@ def hand_pose(
 
             arm_cartesian_command = arm_command_pb2.ArmCartesianCommand.Request(
                 root_frame_name=ref_frame,
+                wrist_tform_tool=geometry_pb2.SE3Pose(
```
amessing-bdai, line 499: @llee-bdai Is this the function that you recently fixed? I am trying to figure out how much of a breaking change this is.

scastro-bdai (commented):
Given that this function is called `hand_pose` but really sets the pose of the arm, and also is called in the `arm_pose_cmd_callback` of the `SpotROS` node...

how would you feel about

1. Renaming this to `set_arm_pose()`
2. Actually making this transform an argument of the function instead of hard-coding, but still defaulting it to identity?
3. Making the above default arg an `SE3` and converting it to the protobuf inside this function body?

khughes-bdai:
@scastro-bdai i am nervous that renaming from `hand_pose()` could be a breaking change to external users -- although there already was a breaking change to this function with https://github.com/bdaiinstitute/spot_wrapper/pull/145 and there were no complaints 😅 
To your third point, we unfortunately can't use spatialmath `SE3` in this repo as it's not a dependency. I think a bosdyn `SE3Pose` could make sense as a default arg

llee-bdai:
@amessing-bdai pulling you into comments here because wer're all talking about the same thing. To answer your question as to if this is the function I recently fixed, yes, in terms of fixing it by removing ROS1 dependencies and actually using it. This isn't a breaking change for anything internal, but it would be breaking externally, technically.

@scastro-bdai @khughes-bdai 
1. I am not opposed to changing the name. The only reason it is named what it is now is because it has been this way since being written. If we were to rename it, I would agree to `set_arm_pose`
2. and 3. I would rather not make this an argument for now until we have a reason to make it an argument. We already have so many args into this function, I would hesitate to add more until we need it.

amessing-bdai:
Hmm, since we recently changed the signature already it's probably ok to further change it, but lets make sure this is the last change to the signature and not make this common practice.

jbarry-bdai (changes requested):
IIRC the offset that it defaults to is the offset from the wrist to the center of the gripper, which is usually the "tool frame" people want.

I think this is probably not a desirable change especially this deep in the code in a repo that lots of people use...

scastro-bdai:
> IIRC the offset that it defaults to is the offset from the wrist to the center of the gripper, which is usually the "tool frame" people want.
> 
> I think this is probably not a desirable change especially this deep in the code in a repo that lots of people use...

I guess as a compromise, should we include the optional offset argument, and default it to the "sanctioned" tool frame?

khughes-bdai:
If you want to be explicitly commanding the wr1 link, it might make sense to make a new function called something like `wrist_pose`. This would eliminate any confusion with the pre-existing `hand_pose` function and make it clear that it's commanding a different link of the robot. thoughts?

llee-bdai:
I'm cool with @khughes-bdai suggestion of adding a whole new wrist pose function. But also, if we add another function, I wouldn't mind just adding in all the extra params required to specify wrist to some offset pose

llee-bdai:
On second thought, because this change ends up affecting a lot more than I expected, I am going to close this. Thanks for all the comments

PR #152, opened 2025-01-26
fix: pinned np to 1.25.2

mschweig:
pip dependency resolver installs numpy >2 with `[install_spot_ros2.sh](https://github.com/bdaiinstitute/spot_ros2/blob/main/install_spot_ros2.sh)`

Conflicts with ROS2 and several packages such as `cv_bridge`

Pinned numpy to 1.25.2 in `requirements.txt`

Fixes:

https://github.com/bdaiinstitute/spot_ros2/issues/492

https://github.com/bdaiinstitute/spot_ros2/issues/555

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,4 +15,4 @@ pytest-cov==4.1.0
 pytest-xdist==3.5.0
 pyyaml>=6.0
 setuptools==59.6.0
-numpy==1.25.2
+numpy==1.21.5
--- a/requirements.txt
+++ b/requirements.txt
@@ -15,3 +15,4 @@ pytest-cov==4.1.0
 pytest-xdist==3.5.0
 pyyaml>=6.0
 setuptools==59.6.0
+numpy==1.25.2


requirements.txt
```diff
@@ -15,3 +15,4 @@ pytest-cov==4.1.0
 pytest-xdist==3.5.0
 pyyaml>=6.0
 setuptools==59.6.0
+numpy==1.25.2
```
mhidalgo-bdai, line 18: @mschweig why 1.25.2? Latest pre-2.0 on PyPI is 1.26.4, and the one in debians for 22.04 is 1.21.5.

Also, missing EOF newline.

mhidalgo-bdai (changes requested):
Thanks for the contribution @mschweig! Mind to sign your commit for the DCO check to pass?

requirements.txt
mschweig: Hi @mhidalgo-bdai, yeah lets keep it consistent with jammy. Pinned it to 1.21.5.

mhidalgo-bdai (approved):
LGTM with green CI

khughes-bdai (approved):
LGTM!

PR #154, opened 2025-02-04
[SW-1890] modify lease acquiring/taking 

tcappellari-bdai:
Add try catch in `getLease()`. Trying to have better error handling and messages when lease acquisition fails


tcappellari-bdai:
no longer necessary due to design change

PR #156, opened 2025-02-12
Fix false trajectory success when blocked by obstacle

gaddison-bdai:
As per [jira ticket](https://theaiinstitute.atlassian.net/browse/SW-2016), we previously denoted “reaching the goal” as being when basic_command_pb2.SE2TrajectoryCommand.Feedback.Status has value STATUS_GOING_TO_GOAL, or STATUS_NEAR_GOAL, both of which, according to spot-sdk,  were “possibly confusing in situations where the robot cannot achieve the requested trajectory and are now deprecated” ([link](https://dev.bostondynamics.com/protos/bosdyn/api/proto_reference.html#se2trajectorycommand-feedback-status)).

I have updated the logic in spot_wrapper to no longer use those deprecated flags. This is necessary for spot_driver in spot_ros2 to properly report when the robot is blocked by an obstacle.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -270,8 +270,10 @@ class AsyncIdle(AsyncPeriodicQuery):
                 )
 
                 if status == basic_command_pb2.SE2TrajectoryCommand.Feedback.STATUS_STOPPED:
-                    self.stopped = True
                     # Robot is stopped
+                    self.stopped = True
+                    self._spot_wrapper._trajectory_status_unknown = False
+                    self._spot_wrapper.last_trajectory_command = None
                     if (
                         final_goal_status
                         == basic_command_pb2.SE2TrajectoryCommand.Feedback.FINAL_GOAL_STATUS_ACHIEVABLE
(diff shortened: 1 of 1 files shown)


spot_wrapper/wrapper.py
```diff
@@ -265,21 +265,40 @@ def _start_query(self) -> None:
...
+                    # Robot is stopped
+                    if (
+                        final_goal_status
+                        == basic_command_pb2.SE2TrajectoryCommand.Feedback.FINAL_GOAL_STATUS_ACHIEVABLE
+                    ):
+                        self._spot_wrapper.trajectory_complete = True
+                        self._spot_wrapper.at_goal = True
+                        # Clear the command once at the goal
+                        self._spot_wrapper.last_trajectory_command = None
+                        self._spot_wrapper._trajectory_status_unknown = False
```
khughes-bdai, line 283: question - should these two be set for every option if `status == STATUS_STOPPED`? Or only for this specific case?

khughes-bdai: also same question for the `trajectory_complete` flag

gaddison-bdai: Good question, I believe it should be set for every STOPPED case except for STOPPED and FinalGoalStatus = FINAL_GOAL_STATUS_IN_PROGRESS, as this situtation may arise if the robot is stopped and blocked, but has not yet ended its trajectory (in case there is a chance it will find a path/a path will open up). Fixed [here](https://github.com/bdaiinstitute/spot_wrapper/pull/156/commits/b12e6bb69fd8b8916b44cab76404e90642d66233).

tcappellari-bdai (approved):
lgtm

khughes-bdai (approved):

PR #157, opened 2025-02-21
Integrate the frame_prefix argument and make robot_name optional (related to https://github.com/bdaiinstitute/spot_ros2/pull/506).

Imaniac230:
This PR is related to the changes proposed in https://github.com/bdaiinstitute/spot_ros2/pull/506.
* The new explicit `frame_prefix` parameter is exposed to the wrapper.
* The `robot_name` parameter is now actually optional.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -372,8 +372,9 @@ class SpotWrapper:
         password: str,
         hostname: str,
         port: int,
-        robot_name: str,
         logger: logging.Logger,
+        robot_name: typing.Optional[str] = None,
+        frame_prefix: typing.Optional[str] = None,
         start_estop: bool = True,
         estop_timeout: float = 9.0,
         rates: typing.Optional[typing.Dict[str, float]] = None,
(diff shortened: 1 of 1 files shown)


spot_wrapper/wrapper.py
```diff
@@ -372,8 +372,9 @@ def __init__(
         password: str,
         hostname: str,
         port: int,
-        robot_name: str,
         logger: logging.Logger,
+        robot_name: typing.Optional[str] = None,
```
khughes-bdai, line 376: I'm good with adding the frame prefix topic, but I'm worried about making the robot name optional as it changes the order of the arguments, and there's some code we have internally that would break with this. How necessary is this for the spot_ros2 PR?

Imaniac230: Shouldn't really be necessary at all. It was more of a unification thing, since I saw that it was already being handled and passed in as an optional:

https://github.com/bdaiinstitute/spot_wrapper/blob/d96cfb8e66cad5791fed8f40dc03bdb83ac3720f/spot_wrapper/wrapper.py#L423-L425

https://github.com/bdaiinstitute/spot_ros2/blob/702c2fa967735f5cde9d7b16937aea0f408ecbbc/spot_driver/spot_driver/spot_ros2.py#L305-L307

https://github.com/bdaiinstitute/spot_ros2/blob/702c2fa967735f5cde9d7b16937aea0f408ecbbc/spot_driver/spot_driver/spot_ros2.py#L367


Only the prefix should be an actual optional, because an empty string prefix is still considered a valid option.

Imaniac230: Reverted the changes made to robot name in https://github.com/bdaiinstitute/spot_wrapper/pull/157/commits/64b9497f08f9d843cad29676090266be957340d8.

khughes-bdai (approved):

PR #160, opened 2025-03-26
Fail standing mock gRPC calls on shutdown

mhidalgo-bdai:
Precisely what the title says. Otherwise, failing to complete a call leads to test hangs. FYI @khughes-bdai.


tcappellari-bdai (approved):

khughes-bdai (approved):
Thank you!

PR #161, opened 2025-04-03
sw-2047 fixed estop bug

gaddison-bdai:
Calling robot's "estop/gentle" or "estop/hard" service would through an "attribute not found" error.

This was fixed by first checking whether this attribute (spot_wrapper's self._estop_keepalive) was non-None, and resetting (and thus instantiating) the estop if not, while throwing a warning to the user that the software estop was not previously initialized. I figured initializing the estop was preferable over simply telling the user the estop did not exist and thus could not be called, for safety reasons. But open to suggestions!

This was tested on Opal.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -997,6 +997,10 @@ class SpotWrapper:
             on the ground first
         """
         try:
+            if self._estop_keepalive is None:
+                self._logger.warning("Software e-stop keepalive does not exist; creating one now and asserting e-stop.")
+                self.resetEStop()
+
             if severe:
                 self._estop_keepalive.stop()
             else:


spot_wrapper/wrapper.py
```diff
@@ -997,6 +997,10 @@ def assertEStop(self, severe: bool = True) -> typing.Tuple[bool, str]:
             on the ground first
         """
         try:
+            if self._estop_keepalive is None:
+                self._logger.warning("Software e-stop keepalive does not exist; creating one now and asserting e-stop.")
+                self.resetEStop()
+
```
khughes-bdai, line 1003: I'm worried this will lead to undesirable behavior if the user does not launch the driver with the estop i.e., if the tablet holds it, this seems like the wrapper will try to take it away. I think in this case we should not create a new estop object, and simply return early (maybe with a logger warning). What do you think?

gaddison-bdai: Do you mean, still try to estop the robot, but do so through a new estop object rather than through the logic in resetEStop()?

gaddison-bdai: @khughes-bdai  While I agree that stealing the estop from the tablet may not sound great, my though process was that we have two options upon receiving a software estop request with no software estop present:

1. we estop anyway, potentially "stealing" it from another device, so as to take the safer option in case the user needed to estop but forgot which endpoint was configured, or
2. we do not estop, and tell the user the estop endpoint is somewhere else, forcing them to potentially have to find the other estop device that they may have forgotten about at this point, all during an estop-worthy event.


Overall, I was thinking false positive estop may be desireable for safety reasons over a false negative. Let me know if this makes sense or if there is larger overall architecture/UX I am omitting.

khughes-bdai (approved):
Sorry for the delayed review!

spot_wrapper/wrapper.py
khughes-bdai: Makes sense, I'm good with this

tcappellari-bdai (approved):

PR #169, opened 2025-06-04
possible lidar fix

JustinBoverhof:
My fix to the issues brought up in [This Issue](https://github.com/bdaiinstitute/spot_ros2/issues/669)

It makes small changes to how the wrapper is handling the Lidar data callbacks. There is also a small change needed in the Driver code.

### DIFF
--- a/spot_wrapper/spot_eap.py
+++ b/spot_wrapper/spot_eap.py
@@ -45,7 +45,8 @@ class AsyncPointCloudService(AsyncPeriodicQuery):
 
         if self._callback is not None and len(self._point_cloud_requests) > 0:
             callback_future = self._client.get_point_cloud_async(self._point_cloud_requests)
-            callback_future.add_done_callback(self._callback)
+            #callback_future.add_done_callback(self._callback)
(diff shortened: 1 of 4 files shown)


spot_wrapper/spot_eap.py
```diff
@@ -45,7 +45,8 @@ def _start_query(self) -> typing.Optional[FutureWrapper]:
 
         if self._callback is not None and len(self._point_cloud_requests) > 0:
             callback_future = self._client.get_point_cloud_async(self._point_cloud_requests)
-            callback_future.add_done_callback(self._callback)
+            #callback_future.add_done_callback(self._callback)
```
tcappellari-bdai, line 48: nit: remove dead code

tcappellari-bdai (commented):
Thanks for the contribution! On first pass (for both PRs) they look good. We will test on hardware on our end just to sanity check before approving and merging

khughes-bdai (approved):
LGTM, just tested on hardware as a sanity check with no issues!

PR #171, opened 2025-06-17
[N/A] fgillespie/Include current obstacle params in velocity_cmd Signed-off-by: Fiona Gillespie <fgillespie@rai-inst.com>

fgillespie-bdai:
This change allows the obstacle avoidance settings to persist through velocity commands rather than being overwritten. This feature is particularly useful for EEP as we allow the public to control Spot in our summer pop-up. 

There is an optional flag for whether to pass through the obstacle params or just use the previous behavior. The flag defaults to False to make sure Spot Wrapper users can continue operation as they expect. 

This change has been tested on Spot.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -1275,7 +1275,13 @@ class SpotWrapper:
         return self._mobility_params
 
     def velocity_cmd(
-        self, v_x: float, v_y: float, v_rot: float, cmd_duration: float = 0.125, body_height: float = 0.0
+        self,
+        v_x: float,
+        v_y: float,
+        v_rot: float,
+        cmd_duration: float = 0.125,
+        body_height: float = 0.0,
+        use_obstacle_params: bool = False,
     ) -> typing.Tuple[bool, str]:
         """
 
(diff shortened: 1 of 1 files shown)


spot_wrapper/wrapper.py
```diff
@@ -1275,7 +1275,13 @@ def get_mobility_params(self) -> spot_command_pb2.MobilityParams:
         return self._mobility_params
 
     def velocity_cmd(
-        self, v_x: float, v_y: float, v_rot: float, cmd_duration: float = 0.125, body_height: float = 0.0
+        self,
+        v_x: float,
+        v_y: float,
+        v_rot: float,
```
fgillespie-bdai, line 1281: test

spot_wrapper/wrapper.py
```diff
@@ -1275,7 +1275,13 @@ def get_mobility_params(self) -> spot_command_pb2.MobilityParams:
...
+        cmd_duration: float = 0.125,
+        body_height: float = 0.0,
+        use_obstacle_params: typing.Optional[bool] = False,
```
khughes-bdai, line 1284: nit: This could just be
```suggestion
        use_obstacle_params: bool = False,
```

khughes-bdai (approved):

PR #174, opened 2025-08-14
implement spot leash resources for missing status/leases

fzillenbiller:
What changed
- Added resources property to SpotLeash returning the latest LeaseResource list from AsyncLease.

Why
- status/leases is not published after the spot leash redesign (worked in 4.0.1) as resources is only defined in SpotLeashProtocol but not implemented in SpotLeash. Implementing it restores the existing publication path in direct leasing mode without changing behavior elsewhere.

How it was tested
- Launched the node with leasing_mode=direct and lease_rate>0.
- Verified that the ROS2 topic status/leases is published (received non-empty LeaseArray).

### DIFF
--- a/spot_wrapper/spot_leash.py
+++ b/spot_wrapper/spot_leash.py
@@ -132,6 +132,10 @@ class SpotLeash(SpotLeashProtocol):
         return self._lease
 
     @property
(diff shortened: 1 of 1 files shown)


tcappellari-bdai (approved):
I was able to test and verify everything works on my end. Thank you for the contribution!

PR #179, opened 2025-09-03
Added timesync endpoint for the arm velocity control 

kabirkedia:
This PR addresses bdaiinstitute/spot_ros2#726.

Changes

Updated the arm velocity command callback to include proper timestamp handling.

Instead of restamping with time.time(), the ROS message’s header timestamp is now passed directly into the SDK.

Why this matters

Preserving the original ROS timestamp ensures that the commands reflect the actual time they were generated rather than the time they happened to be processed. This is especially important for teleop scenarios, where even small timing inconsistencies can affect responsiveness and control smoothness.

### DIFF
--- a/spot_wrapper/wrapper.py
+++ b/spot_wrapper/wrapper.py
@@ -1290,7 +1290,7 @@ class SpotWrapper:
         v_x: float,
         v_y: float,
         v_rot: float,
-        timestamp: float = now_sec(),
+        timestamp: float | None = None,
         cmd_duration: float = 0.125,
         body_height: float = 0.0,
         use_obstacle_params: bool = False,
@@ -1312,7 +1312,8 @@ class SpotWrapper:
         Returns:
             Tuple of bool success and a string message
         """
-        end_time = timestamp + cmd_duration
+        start_time = now_sec() if timestamp is None else timestamp
+        end_time = start_time + cmd_duration
         if body_height:
             current_mobility_params = self.get_mobility_params()
             height_adjusted_params = RobotCommandBuilder.mobility_params(


spot_wrapper/wrapper.py
```diff
@@ -1289,6 +1289,7 @@ def velocity_cmd(
         v_x: float,
         v_y: float,
         v_rot: float,
+        timestamp: float,
```
khughes-bdai, line 1292: I would prefer to make this an optional field that defaults to None so it does not break the existing API, if it's not set it can default to time.time() in the function body? 
Could you also add a description of this new arg in the docstring?

spot_wrapper/spot_arm.py
```diff
@@ -539,14 +539,17 @@ def handle_arm_velocity(
                 self._logger.info(msg)
                 return False, msg
             else:
-                end_time = self._robot.time_sync.robot_timestamp_from_local_secs(time.time() + cmd_duration)
-                arm_velocity_command.end_time.CopyFrom(end_time)
+                end_time = arm_velocity_command.end_time.seconds + arm_velocity_command.end_time.nanos/1e9 + cmd_duration
```
khughes-bdai, line 542: Is this the expected behavior? It seems like this will make the command end at `end_time + cmd_duration` instead of just `end_time`.

kabirkedia: Ohh yes!
I wasn't really sure on how to proceed.
I don't pass an end time in my header. What would you prefer?

kabirkedia: I just use the end time field as a current timestamp, I wasn't really sure on how to add it. 
What is your opinion?

khughes-bdai: I think it is a little confusing to modify the end time field from the command as this has a specific meaning to the Spot SDK. I think you could also get into a situation where if you forget to specify this when you are constructing the robot command, the end time defaults to 0.0 and your command will never get processed since this time is in the past. 

I think that combined with the QOS change that you are planning to introduce upstream in spot_ros2, it should be enough to just apply the change from your first commit here (https://github.com/bdaiinstitute/spot_wrapper/pull/179/commits/e4cefa1777d6947b0b8e65fd31c95d4c07b18da5) which does use time.time() and passes the endpoint appropriately. What do you think? 

(If you want to be really explicit about forcing the user to pass a timestamp, I think a solution could be to make a custom message that explicitly includes a timestamp/header field, but think that might be overkill in addition to the QOS change.)

kabirkedia:
Hey @khughes-bdai @tcappellari-bdai did you have a chance to check?

kabirkedia:
Okay thank you!

khughes-bdai:
Hey @kabirkedia, so sorry for the delay in the review. First thing, there have been a few changes to `spot_wrapper` coming from https://github.com/bdaiinstitute/spot_wrapper/pull/182 that I think are now causing merge conflicts with this PR. That PR switches to using BD's `now_sec()` instead of `time.time()` which for the purposes of this PR are equivalent. 

My opinion on which way to proceed with this and https://github.com/bdaiinstitute/spot_ros2/pull/734 would be to try the QOS change on the spot ros2 side, and on this side just pass the endpoint appropriately to the robot command. This to me seems like the simplest way to proceed. If this is not enough, then we can evaluate other options like explicitly passing timestamps. 

Thanks again for all this work, I think it will be super useful to get in!

kabirkedia:
Hey @khughes-bdai I think I made the correct merge. Please check and let me know!

kabirkedia:
Hey @khughes-bdai I ran pre-commit on this
Let me know if this still works

PR #183, opened 2025-09-24
[SW-3698] Send spot hand calibration from yaml

tcappellari-bdai:
Added the ability to send spot new hand camera calibrations from a previously computed yaml file. Just add the `-yaml` or `--from_yaml` flag when running the calibration script and it will use your provided data path to the yaml (include data path with `-dp` flag)


mvictora-bdai (commented):
Looks good to me, minus a leftover unused commented line of code

khughes-bdai (approved):
+1 to michelle's comment but otherwise LGTM!

PR #184, opened 2025-10-14
Update aiortc version

khughes-bdai:
Updates the aiortc version for this package. This package is used for spot cam dependencies. 

## Tested
- [ ] CI


khughes-bdai:
@mhidalgo-bdai yeah, we were already ahead of the desired version :( I bumped it because we were having mass issues with lower versions of aiortc causing CI failures (both in this repo and other repos that pull this one in). If it causes issues we can revert and come up with a better solution.

PR #189, opened 2025-12-19
[SW-4178] Spot calibration params bug

tcappellari-bdai:
the values sent to Spot to override its default calibration for the depth and rbg cameras was incorrect


khughes-bdai (approved):
LGTM, might be good to check this with @lgalup-bdai since i know he's been trying to debug calibration issues lately too

PR #193, opened 2026-03-05
Fix calibration script

tcappellari-bdai:
Fixes the calibration script (originally created by Gary) after a bug was discovered. Much of the fixes are based off of Lu's work in `bdai_calibration`. Also updates the readme 

Tested on Jasper and Ishmael (their internal gripper camera params are set to the newly calculated ones from testing)

Screenshot from 2026-03-05 13-33-37


spot_wrapper/calibration/spot_in_hand_camera_calibration.py
```diff
@@ -181,11 +181,17 @@ def convert_pinhole_intrinsic_to_proto(intrinsic_matrix: np.ndarray) -> ImageSou
...
 
-        depth_to_planning = cal["depth_to_hand"] @ hand_t_planning
-        rgb_to_planning = rgb_t_depth @ depth_to_planning
+        # cal["depth_to_hand"] is rgb_M_hand: the hand frame expressed in RGB camera coords.
+        # hand_t_planning is hand_M_wr1 (hand->wr1 transform built above).
+        # rgb_M_wr1 = rgb_M_hand @ hand_M_wr1
+        rgb_M_wr1 = cal["depth_to_hand"] @ hand_t_planning
+        # rgb_t_depth is depth_M_rgb (stereo extrinsic: maps RGB points → depth frame).
+        # depth_M_wr1 = depth_M_rgb @ rgb_M_wr1
+        depth_M_wr1 = rgb_t_depth @ rgb_M_wr1
```
mhidalgo-rai, line 190: @tcappellari-bdai meta: IIUC `rgb_t_depth` is in fact `depth_M_rgb`, so this isn't a bug. Still, even if the calibration dictionary uses a different convention, we should make sure ours aligns.

mhidalgo-rai (commented):
This a bit of a large patch and I have limited context. What is this fixing exactly?

khughes-bdai (approved):
I haven't reviewed thoroughly but I know you and Lu have been working on this for a while. If it's working on robot and @lgalup-bdai is good with the underlying math, I'll stamp it

lgalup-bdai:
folks, i cant visually verify if the math is correct. that would be like visually verifying that software has no bugs. is there something we can do to test it?

spot_wrapper/calibration/spot_in_hand_camera_calibration.py
lgalup-bdai: agreed. its very confusing to use both types of notation.

khughes-bdai:
looks like @tcappellari-bdai has run it on robot -- could you see an improvement with this vs the old method?

tcappellari-bdai:
@khughes-bdai Ya so previously, when i saved the new calibration params to the robot (after fixing the initial frame transform errors), the depth hand camera would just stop working. Like just sending empty messages and a black screen on the tablet. The calibration was also saying the depth and RBG cameras in the hand were like 5 cm apart (way too big). After these changes, the cameras are both fully functional and the frame positions and orientations look correct in rviz. 

(also Josh ran a spot check on Jasper after I changed the params and it passed all the checks, for whatever thats worth)

spot_wrapper/calibration/spot_in_hand_camera_calibration.py
tcappellari-bdai: I had to use claude for help debugging and it seems to prefer the other convention :smiling_face_with_tear:

